PR #28, opened 2018-10-18
Feature/urdf examples

romanarust:
Hi Kathrin and Gonzalo,

Mainly added 3 examples on urdf/moveit, that need to be checked on
* english
* comprehensibleness
* ...

I also made some changes on robot, artist, rosfileserverloader (wip) which should have move into a different branch (sorry)..


src/compas_fab/backends/ros/messages/services.py
```diff
@@ -164,4 +166,52 @@ def __init__(self, planner_config='', group='', params=PlannerParams(), replace=
...
+    @property
+    def msg(self):
+        msg = super(MotionPlanRequest, self).msg
+        return {"motion_plan_request":msg}
+
+class MotionPlanResponse(ROSmsg):
+    """http://docs.ros.org/kinetic/api/moveit_msgs/html/msg/MotionPlanResponse.html
+    """
+
+    def __init__(self, trajectory_start=RobotState(), group_name="",
```
gonzalocasas: Isn't it problematic to initialize like this? I think they all share the same instance across the entire program lifetime. It's safer to default to `None` and initialize inside the constructor to the specific default, e.g. `RobotState()`.
Unless these defaults are immutable, then it's not a problem to keep as is.

src/compas_fab/robots/robot.py
```diff
@@ -30,8 +33,10 @@ class Robot(object):
 
     Attributes
     ----------
-    robot_model : :class:`compas.robots.Robot`
+    model : :class:`compas.robots.Robot`
         The robot model, usually created out of an URDF structure.
+    artist : :class:`compas_fab.artists.BaseRobotArtist`
+        TODO
```
gonzalocasas:
```suggestion
        Instance of the artist used to visualize the robot.
```

src/compas_fab/robots/robot.py
```diff
@@ -83,12 +89,15 @@ def from_resource_path(cls, directory, client=None):
...
-        urdf_importer = UrdfImporter.from_robot_resource_path(directory)
+        urdf_importer = RosFileServerLoader.from_robot_resource_path(directory)
         urdf_file = urdf_importer.urdf_filename
+        
         srdf_file = urdf_importer.srdf_filename
         urdf_model = compas.robots.model.Robot.from_urdf_file(urdf_file)
         srdf_model = RobotSemantics.from_srdf_file(srdf_file, urdf_model)
-        return cls(urdf_model, srdf_model, client)
+        print("urdf_model", urdf_model)
+        print("srdf_model", srdf_model)
```
gonzalocasas: Left over `print` debug statements. If intended to stay, turn them into log statements instead (but probably too verbose).

docs/examples/07_ros_create_urdf_ur5_with_measurement_tool.rst
```diff
@@ -0,0 +1,325 @@
...
+  <exec_depend>joint_state_publisher</exec_depend>
+  <exec_depend>robot_state_publisher</exec_depend>
+  <exec_depend>rviz</exec_depend>
+  <exec_depend>xacro</exec_depend>
+
+Optionally, modify ``email`` and ``licence``, ``version`` tags.
+
+Then create 4(+2) folders: ``launch``, ``rviz``, ``urdf`` and ``meshes`` (with visual and collision folders)::
+
+  mkdir launch
```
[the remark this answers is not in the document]
romanarust: cool! I did not know that! ;)

docs/examples/09_ros_create_urdf_ur10_on_tower.rst
```diff
@@ -0,0 +1,393 @@
+.. _ros_examples:
+
+********************************************************************************
+Creating a URDF of the UR10 on the tower
+********************************************************************************
+
+**ATTENTION:**
```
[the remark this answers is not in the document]
romanarust: Yes, it refers to an equipment we own, that is why it has to be checked before using ;), but could anyway serve as an example of a robot on linear axes. I'll leave it there, but remove the attention note...

gonzalocasas:
💃

PR #31, opened 2019-01-15
Feature/more moveit functionality

romanarust:
Added more moveit functionality to the library: add `CollisionObject` to scene, add collision object to end-effector (`AttachedCollisionObject`). 
Added also "pre"-callbacks in compas_fab robot to convert parts of the Moveit response message to `Configuration` or `Frame`. Not sure if this is the right way or we start implementing compas_fab Response types...


romanarust:
Thank U

PR #40, opened 2019-04-22
Urdf load sample

gonzalocasas:
This PR is a refactoring of the ROS File Server Loader to move it inside the `ros` backend package and align it to have an API closer to that of the other loaders:

* Move d`RosFileServerLoader` to `compas_fab.backends.ros` package (and importable from `compas_fab.backends` together with the `RosClient`)
* Changed the loading logic to be synchronous
* Added a bunch of examples to documentation of different loaders
* As an almost side-effect, implemented a WIP DAE importer. It is right now a very quick hack thing, but maybe the basis of something more elaborate.


src/compas_fab/backends/ros/fileserver_loader.py
```diff
@@ -0,0 +1,303 @@
...
+            _, local_filename = tempfile.mkstemp(suffix='.' + file_extension, prefix='ros_fileserver_')
+
+        if not use_local_file:
+            service = roslibpy.Service(self.ros, '/file_server/get_file', 'file_server/GetBinaryFile')
+            request = roslibpy.ServiceRequest(dict(name=url))
+            response = await_callback(service.call, 'callback', 'errback', request)
+
+            file_content = binascii.a2b_base64(response.data['value'])
+
+            # Just look away, we're about to do something nasty!
```
romanarust: haha ;)

romanarust:
Great! Finally, the ROS file server loader made it to the documentation ;)
I like the loading documentation page.

PR #43, opened 2019-04-25
improve comments for better doc.

xarthurx:
Things that are not clear based on the current doc and figured out from asking questions.

Hoping to create a better doc little by little.


gonzalocasas:
👍

PR #47, opened 2019-05-28
Pose class

matteo-pacher:
Pose class to handle:
- different quaternion conventions
- quaternion canonization (scalar part of the quaternion always >=0)
- unitized quaternion (only unit quaternions describe an orientation)
- add rotational deltas
- quaternion to and from ABB euler angles


romanarust:
Hi Matteo, nice contribution. 
But a `Pose` class seems redundant if there is already the `Frame`, which has the same information stored. Rather, it would be cool if you can turn this into a `Quaternion` class.

gonzalocasas:
@funkchaser just talking with @matteo-pacher about this, and it seems we can delete this branch since the quaternion feature is now part of `COMPAS`. Do you agree?

romanarust:
closing this, since quaternion now implemented in compas

PR #67, opened 2019-09-04
Update `compas_fab.artists.BaseRobotArtist` and `compas.robots.RobotModel`

romanarust:
Several smaller changes, mainly moving functionality from `compas_fab.artists.BaseRobotArtist` to `compas.robots.RobotModel`. So this pull request is linked with another one.
The main advantage is to call forward_kinematics on the robot model and to have a bit of better divide between artist and model.
These are breaking, but easily fixable changes.

### DIFF
--- a/src/compas_fab/artists/base.py
+++ b/src/compas_fab/artists/base.py
@@ -59,8 +59,8 @@ class BaseRobotArtist(object):
 
         Parameters
         ----------
-        geometry : :class:`compas.datastructures.Mesh`
-            Instance of a **COMPAS** mesh.
+        geometry : :class:`compas.datastructures.Mesh` or :class:`compas.geometry.Shape`
+            Instance of a **COMPAS** mesh or **COMPAS** shape
 
         Returns
         -------
--- a/src/compas_fab/artists/base.py
+++ b/src/compas_fab/artists/base.py
@@ -151,7 +151,7 @@ class BaseRobotArtist(object):
         self.transform(item.native_geometry, relative_transformation)
         item.current_transformation = transformation
 
-    def update(self, configuration, names, collision=True):
+    def update(self, configuration, names, visual=True, collision=True):
         """Triggers the update of the robot geometry.
 
         Parameters
@@ -160,6 +160,9 @@ class BaseRobotArtist(object):
             Instance of the configuration (joint state) to move to.
         names : list of string
             The names of the configurable joints to update.
+        visual : bool, optional
+            ``True`` if the visual geometry should be also updated, otherwise ``False``.
+            Defaults to ``True``.
         collision : bool, optional
             ``True`` if the collision geometry should be also updated, otherwise ``False``.
             Defaults to ``True``.
--- a/src/compas_fab/artists/base.py
+++ b/src/compas_fab/artists/base.py
@@ -23,7 +23,7 @@ class BaseRobotArtist(object):
 
     There are two methods that implementors of this base class should provide, one
     is concerned with the actual creation of geometry in the native format of the
-    CAD environment (:meth:`draw_mesh`) and the other is one to apply a transformation
+    CAD environment (:meth:`draw_geometry`) and the other is one to apply a transformation
     to geometry (:meth:`transform`).
 
     Attributes
@@ -35,23 +35,23 @@ class BaseRobotArtist(object):
     def __init__(self, robot):
         super(BaseRobotArtist, self).__init__()
         self.robot = robot
-        self.create(robot.root, Transformation())
+        self.create()
         self.scale_factor = 1.
 
-    def transform(self, native_mesh, transformation):
-        """Transforms a CAD-specific mesh using a **COMPAS** transformation.
+    def transform(self, geometry, transformation):
+        """Transforms a CAD-specific geometry using a **COMPAS** transformation.
 
         Parameters
         ----------
-        native_mesh : object
-            A CAD-specific (i.e. native) geometry object as returned by :meth:`draw_mesh`.
+        geometry : object
+            A CAD-specific (i.e. native) geometry object as returned by :meth:`draw_geometry`.
         transformation : `Transformation`
             **COMPAS** transformation to update the geometry object.
         """
         raise NotImplementedError
 
-    def draw_mesh(self, mesh, color=None):
-        """Draw a **COMPAS** mesh into the CAD environment.
+    def draw_geometry(self, geometry, color=None):
+        """Draw a **COMPAS** geometry in the respective CAD environment.
 
         Note
         ----
@@ -59,7 +59,7 @@ class BaseRobotArtist(object):
 
         Parameters
         ----------
-        mesh : :class:`compas.datastructures.Mesh`
+        geometry : :class:`compas.datastructures.Mesh`
             Instance of a **COMPAS** mesh.
 
         Returns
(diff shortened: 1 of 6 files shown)


src/compas_fab/artists/base.py
```diff
@@ -35,31 +35,31 @@ class BaseRobotArtist(object):
...
+        """Draw a **COMPAS** geometry in the respective CAD environment.
 
         Note
         ----
         This is an abstract method that needs to be implemented by derived classes.
 
         Parameters
         ----------
-        mesh : :class:`compas.datastructures.Mesh`
+        geometry : :class:`compas.datastructures.Mesh`
```
gonzalocasas: If the name changed to `draw_geometry`, does that mean we should change the input data type from `compas.datastructures.Mesh` to something else? (although, there's no root class for compas geometry)

src/compas_fab/ghpython/artists.py
```diff
@@ -24,7 +24,7 @@ def transform(self, native_mesh, transformation):
         T = xform_from_transformation(transformation)
         native_mesh.Transform(T)
 
-    def draw_mesh(self, compas_mesh, color=None):
+    def draw_geometry(self, compas_mesh, color=None):
```
gonzalocasas: same comment, maybe instead of `compas_mesh` this should be called something else if it accepts more than just meshes?

src/compas_fab/robots/robot.py
```diff
@@ -46,11 +46,21 @@ class Robot(object):
     """
 
     def __init__(self, model, artist=None, semantics=None, client=None):
+        self._scale_factor = 1.
         self.model = model
-        self.artist = artist
+        self._artist = None
```
gonzalocasas: I don't think you need to explicitly set it to `None` if the next line uses the property setter that assigns it anyway.

src/compas_fab/robots/robot.py
```diff
@@ -46,11 +46,21 @@ class Robot(object):
...
+        self.artist = artist  # setter and getter (because of scale)
         self.semantics = semantics
         self.client = client  # setter and getter ?
-        self.scale(1.)
+
+    @property
+    def artist(self):
```
gonzalocasas: Although the name is very clear, since this property is important, please add a little docstring to explain it.

src/compas_fab/robots/robot.py
```diff
@@ -360,6 +370,22 @@ def init_configuration(self, group=None):
         positions = [0.] * len(types)
         return Configuration(positions, types)
 
+    def random_configuration(self, group=None):
+        """Returns a random configuration.
+
+        Note that no collision checking is involved, so the configuration may be invalid.
+        """
+        import random
```
gonzalocasas: please move the import to top ;)

src/compas_fab/robots/robot.py
```diff
@@ -1125,25 +1159,27 @@ def send_trajectory(self):
...
     # drawing
     # ==========================================================================
 
-    def update(self, configuration, group=None, visual=True, collision=True):
+    def update(self, configuration, group=None, collision=True):
         """Updates the robot's geometry.
         """
         names = self.get_configurable_joint_names(group)
-        self.artist.update(configuration, names, visual, collision)
+        self.artist.update(configuration, names, collision)
```
gonzalocasas: Wouldn't it be actually desirable to be able to not transform visual geometry very often?

src/compas_fab/robots/robot.py
```diff
@@ -864,15 +873,40 @@ def forward_kinematics(self, configuration, group=None):
...
-        # TODO implement response types
-
         frame_RCF = response.pose_stamped[0].pose.frame
         frame_RCF.point *= self.scale_factor
         response.frame_RCF = frame_RCF
         response.frame_WCF = self.represent_frame_in_WCF(frame_RCF, group)
 
         return response
 
+    def forward_kinematics_robot_model(self, configuration, group=None, link_name=None):
```
gonzalocasas: This name makes me feel something is not right. It's too long, and redundant in the sense that there is a `self.model` but still we need to say `forward_kinematics_robot_model`.

One option would be to change the `forward_kinematics` method, adding an optional parameter and depending on its value, use the `self.model` or use `self.client.forward_kinematics` method.

The name and possible values of this parameter probably need some thought or discussion, but one option would be to call it `backend`. Then the cases to handle are:
* `backend=None and self.client != None` => use client service
* `backend=None and self.client = None` => use local robot model resolution
* `backend='local'` => use local robot model resolution
* `backend != None and backend != 'local'` => use client service and pass that value down to the `self.client.forward_kinematics` method, because the client might do some further selection (e.g. KDL vs UR kinematic model, etc)

If this proves to look like a nice API, we should extend it to the IK side.

src/compas_fab/robots/robot.py
```diff
@@ -1125,25 +1159,27 @@ def send_trajectory(self):
         self.ensure_client()
         raise NotImplementedError
 
-    @property
-    def frames(self):
-        """The robot's frames."""
-        return self.model.frames
+    def updated_frames(self, configuration, group=None):
```
gonzalocasas: Is it really `updated_frames` the correct name for this? Shouldn't it be `transformed_frames`? Or `current_frames` even? Not entirely sure...

src/compas_fab/artists/base.py
```diff
@@ -129,88 +130,54 @@ def scale_link(self, link, transformation):
...
         """
-        positions = configuration.values
-        self.update_links(names, positions, visual, collision)
+        relative_transformation = transformation * item.current_transformation.inverse()
+        self.transform(item.native_geometry, relative_transformation)
+        item.current_transformation = transformation
 
-    def update_links(self, names, positions, visual=True, collision=True):
-        """Updates the joints and link geometries.
+    def update(self, configuration, names, collision=True):
```
gonzalocasas: One more comment: adding the `names` parameter will feels slightly odd here. I assume this is to support partial configurations, but in general, we do this by passing a `group` parameter if the values only apply to a sub-set of the joints.

Could we replace it with `group`, and then if `group is None` we set `names` as default to `robot.get_configurable_joint_names()`?

src/compas_fab/artists/base.py
romanarust: Well, actually it should support primitives such as Box, Cylinder, Sphere, Capsule (yet to be implemented in compas.primitives) and Mesh. 
@brgcode: Here it would make sense to have a Base class for the geometry.primitives then we could use either `compas.datastructures.Mesh` or `compas.geometry.Primitive` as input data type.

src/compas_fab/artists/base.py
romanarust: This is the problem of having a divide between robot semantics and robot model. Only semantics knows about the joint names that belong to the planning group. So they are now passed from the `compas_fab.robots.Robot` over the artist to the model. 
One option would be to move semantics as attribute to the robot model (as discussed once....)

src/compas_fab/artists/base.py
romanarust: changed to `compas.datastructures.Mesh` or `compas.datastructures.Shape` in description

src/compas_fab/ghpython/artists.py
romanarust: changed to `compas.datastructures.Mesh` or `compas.datastructures.Shape` in description

src/compas_fab/robots/robot.py
romanarust: added parameter visual

src/compas_fab/robots/robot.py
romanarust: selected `transformed_frames`

src/compas_fab/artists/base.py
romanarust: After discussion, made a card:  https://github.com/compas-dev/compas_fab/projects/1#card-27015436 to be handled in another PR

romanarust:
changes implemented, ready to merge?

gonzalocasas:
image

PR #75, opened 2019-10-15
Handle generic trajectory arguments on the ROS client trajectory follower

gonzalocasas:
Changed the ROS client trajectory follower to handle (in a backwards-compatible way) arguments of generic trajectory type.

### Checklist

1. [x] Add the change to the `CHANGELOG.rst` file in the `Unreleased` section under the most fitting heading: `Added`, `Changed`, `Removed`.
1. [x] Run all tests on your computer (i.e. `invoke test`).
1. [ ] If you add new functions/classes, check that:
   1. [ ] Are available on a second-level import, e.g. `compas_fab.robots.Configuration`.
   1. [ ] Add unit tests (especially important for algorithm implementations).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ matrix:
       env: TRAVIS_PYTHON_VERSION=3.7
 
 install:
+  - python -m ensurepip
   - pip install cython --install-option="--no-cython-compile"
   - pip install --no-cache-dir -r requirements-dev.txt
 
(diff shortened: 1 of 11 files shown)


gonzalocasas:
After debugging the issue mentioned by @xarthurx above, he confirmed it works (issue was unrelated).

However, I've updated the description of this PR because the change is now a breaking change, the return type of execution methods has been renamed from `CancellableTask` to `CancellableFutureResult`. It is only a name change thou, the method names and semantics of the class remain the same.

@romanarust: considering that, are you still ok with merging it?

romanarust:
of course ;)

gonzalocasas:
merci! then, merging!

PR #89, opened 2019-11-18
Feature: set_end_effector for `Robot`

### LINKED ISSUE #66: Setting a custom tool (not changing URDF)
For robot's end-effectors without moveable parts, the recommended approach is to use specify a collision object rather than changing the URDF (extending by one link). This means that all planning has to be performed with an attached_collison_object and frames (ik, cartesian_motion) have to be calculated with the transformation of the tool tcp to tool0. 
It would be practical if the `compas_fab.robots.Robot` would handle this:, i.e. setting a custom end-effector, and automate the steps needed for the attached_collison_object and frame calculation.

romanarust:
This PR adds the possibility of setting a custom tool (not changing URDF) to the robot. Additionally adding a calculation for tcf and t0cf frames (tool tip and flange).
Closes #66

### Checklist

- [ ] I ran lint on my computer and there are no errors (i.e. `invoke lint`) .


src/compas_fab/robots/robot.py
```diff
@@ -675,6 +681,108 @@ def to_world_coords(self, frame_RCF, group=None):
...
+        Frame(Point(-0.309, -0.046, -0.266), Vector(0.276, 0.926, -0.256), Vector(0.879, -0.136, 0.456))
+        """
+        # TODO: or, rather than raising exception: do we simply read the end-effector from the urdf
+        if not self.end_effector_frame:
+            raise Exception("Please set an endeffector")
+        Te = Transformation.from_frame_to_frame(Frame.worldXY(), self.end_effector_frame)
+        Tc = Transformation.from_frame(frame_t0cf)
+        return Frame.from_transformation(Tc * Te)
+
+    def set_end_effector(self, mesh, frame, group=None, touch_links=None):
```
gonzalocasas: Given the previous suggestions, I would also suggest renaming this to `attach_tool` and change it to receive an instance of `Tool`.

```suggestion
    def attach_tool(self, tool, group=None, touch_links=None):
```

src/compas_fab/robots/robot.py
```diff
@@ -675,6 +681,108 @@ def to_world_coords(self, frame_RCF, group=None):
...
+        """Set's the robots end-effector that is not defined through URDF.
```
gonzalocasas:
```suggestion
        """Attach a tool to the robot independently of the model definition.
```

src/compas_fab/robots/robot.py
```diff
@@ -1333,6 +1453,19 @@ def draw(self):
         """
         return self.draw_visual()
 
+    def draw_end_effector(self, tool0_frame):
```
gonzalocasas: I believe this method should only be dispatching to the artist (also, why does it need `tool0_frame` to be specified?), e.g.:

```python
def draw_attached_tool(self):
    self.artist.draw_attached_tool()
```

and then on the `BaseRobotArtist`, the mesh transformation can be done with native transformations with something like:
```python
def draw_attached_tool(self):
    if not self.robot.attached_tool:
        return

    native_geometry = self.draw_geometry(sefl.robot.attached_tool.mesh)
    # here get the correct frame for the transformation
    # I would expect this to be `self.robot.get_end_effector_frame`, but I'm not sure why instead you had `tool0_frame` here.
    T = Transformation.from_frame_to_frame(self.robot.attached_tool.frame, self.robot.get_end_effector_frame())
    self.transform(native_geometry, T)
    return native_geometry
```

gonzalocasas: Q: Should the artist or the robot also handle scaling of the attached tool?

romanarust:
implemented all requested changes...

src/compas_fab/robots/robot.py
```diff
@@ -10,13 +10,18 @@
 from compas.geometry import Transformation
 from compas.robots import Joint
 from compas.robots import RobotModel
+from compas.datastructures import mesh_transformed
```
gonzalocasas: I think you can remove this import now

CHANGELOG.rst
```diff
@@ -7,16 +7,19 @@ All notable changes to this project will be documented in this file.
...
 **Added**
 
-* ..
+* Added ``attach_tool``, ``detach_tool``, ``draw_attached_tool``, ``to_tool_frame(s)`` and ``to_tool0_frame(s)`` to ``Robot``
+* Added ``attach_tool``, ``detach_tool``, ``draw_attached_tool`` to ``Artist``
+* Added ``add_attached_tool`` and ``remove_attached_tool`` to ``PlanningScene``
 
 **Changed**
 
+* Changed `inverse_kinematics`, `plan_cartesian_motion` and `plan_motion` to use the attached_tool's `AttachedCollisionMesh` if set
```
gonzalocasas:
```suggestion
* Changed ``inverse_kinematics``, ``plan_cartesian_motion`` and ``plan_motion`` to use the attached_tool's ``AttachedCollisionMesh`` if set
```

PR #111, opened 2020-01-25
Suggestion to add option to run pytest-cov through invoke test

tetov:
I though coverage analysis would be helpful for assessing the state of our documentation. This adds a switch to `invoke test` to run the plugin [pytest-cov](https://github.com/pytest-dev/pytest-cov) when testing.

I also switched the default value for the keyword argument `doctest` to `False` to match the command argument help. The Travis build on Linux runs doctest using `pytest --doctest-modules` so I don't think the CI setup will be impacted by this change

```
$ inv test --help
Usage: inv[oke] [--core-opts] test [--options] [other tasks here ...]

Docstring:
  Run all tests.

Options:
  -c, --coverage   True to generate coverage report using pytest-cov
  -d, --doctest    True to run doctest modules, otherwise False.
```

### Checklist
- ~~[ ] I added a line to the `CHANGELOG.rst` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).~~
- ~~[ ] I have added tests that prove my fix is effective or that my feature works.~~

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -60,7 +60,7 @@ script:
   - if [[ "$TRAVIS_PYTHON_VERSION" == "3.7" ]]; then
       python3 -m invoke lint || python -m invoke lint;
     fi
-  - python3 -m invoke test --no-doctest || python -m invoke test --no-doctest
+  - python3 -m invoke test || python -m invoke test
   - if [[ "$TRAVIS_GENERATE_DOCS" == "true" ]]; then
       python3 -m invoke docs || python -m invoke docs;
     fi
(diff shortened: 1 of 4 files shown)


gonzalocasas:
@tetov could you check the failing travis thing? I believe that's because of the `doctest` arg of `tasks.py` changing to `False`; the travis file does some stuff to the pytest options based on the build, so, I assume that now the default is different, some of those don't match.

tetov:
I'll look into it, I'm used to get an email for failing builds but haven't been able to set that up yet, so sorry for not fixing it earlier!

tetov:
Ok, the `--no-doctest` flag is removed from the script part of the travis configuration, meaning it will run without doctests now since that's the new default. However the linux job has `PYTEST_ADDOPTS=--doctest-modules` in env so it runs for that job.

PR #115, opened 2020-02-06
Update copyright year

tetov:


gonzalocasas:
Mmm..no, I disagree with this one, sorry.

> The general rule is that the year to include in a copyright notice is the year of first publication of the work. 

https://www.copyrightlaws.com/copyright-symbol-notice-year/

tetov:
Aha! Didn’t know that!

PR #133, opened 2020-04-02
Configurations search features

matteo-pacher:
Added some functionalities to :class:``compas_fab.robots.Robot``:
- ``position_constraint_from_max_reach`` to constrain a specific link within a maximum distance from a point
- ``iter_inverse_kinematics`` to search for a list of configurations, with given constraints (goal states search for planning goal constraints) 
- ``get_links_distance`` to calculate the distance between two robot's links

- [X] New feature in a **backwards-compatible** manner.

### Checklist
- [X] I added a line to the `CHANGELOG.rst` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [X] I have added tests that prove my fix is effective or that my feature works.
- [X] I have added necessary documentation (if appropriate)

### DIFF
--- a/src/compas_fab/robots/robot.py
+++ b/src/compas_fab/robots/robot.py
@@ -1177,9 +1177,9 @@ class Robot(object):
                                                         constraints=constraints,
                                                         attempts=attempts,
                                                         attached_collision_meshes=attached_collision_meshes,
-                                                        full_joint_state=full_joint_state)
+                                                        return_full_configuration=full_joint_state)
                 result_configurations.append(configuration)
-            except BaseException:
+            except:
                 # print("Error code: -31; NO_IK_SOLUTION")
                 pass
 
--- a/src/compas_fab/robots/robot.py
+++ b/src/compas_fab/robots/robot.py
@@ -4,6 +4,8 @@ from __future__ import print_function
 
 import logging
 import random
+from datetime import datetime
+from datetime import timedelta
 
 from compas.geometry import Frame
 from compas.geometry import Scale
@@ -13,11 +15,11 @@ from compas.robots import Joint
 from compas.robots import RobotModel
 
 from compas_fab.robots.configuration import Configuration
+from compas_fab.robots.constraints import BoundingVolume
 from compas_fab.robots.constraints import Constraint
 from compas_fab.robots.constraints import JointConstraint
 from compas_fab.robots.constraints import OrientationConstraint
 from compas_fab.robots.constraints import PositionConstraint
-
 from compas_fab.robots.planning_scene import AttachedCollisionMesh
 
 LOGGER = logging.getLogger('compas_fab.robots.robot')
@@ -400,7 +402,7 @@ class Robot(object):
         """
         configurable_joints = self.get_configurable_joints(group)
         return [j.type for j in configurable_joints]
-    
+
     def get_links_distance(self, link_name_1, link_name_2, group=None):
         return None
 
@@ -834,7 +836,7 @@ class Robot(object):
         ee_link = self.get_end_effector_link_name(group)
         sphere = Sphere(frame_WCF.point, tolerance_position)
         return PositionConstraint.from_sphere(ee_link, sphere)
-    
+
     def position_constraint_from_max_reach(self, target_point, link_name, max_reach, group=None):
         """
         Returns a position constraint for link_name with a sphere around the target point.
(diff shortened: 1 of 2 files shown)


matteo-pacher:
All the changes for this specific PR are in ``src/compas_fab/robots/robot.py``, all the other edits are inherited by the previous `start FULL configuration for planning service calls` PR

src/compas_fab/robots/robot.py
```diff
@@ -911,6 +864,27 @@ def position_constraint_from_frame(self, frame_WCF, tolerance_position, group=No
         sphere = Sphere(frame_WCF.point, tolerance_position)
         return PositionConstraint.from_sphere(ee_link, sphere)
 
+    def position_constraint_from_max_reach(self, target_point, link_name, max_reach, group=None):
```
romanarust, line 867: I don't get the added value of this function. The only difference to `position_constraint_from_frame` is that you can pass the link name. If that is desired, than I'd suggest to rather add an optional parameter to `position_constraint_from_frame`.
Additionally, the scaling happens only before the planning request and should stay there.

If you found out that the way the scaling on a `Sphere` is wrong, please fix or raise an issue on COMPAS.

src/compas_fab/robots/robot.py
```diff
@@ -1076,30 +1050,141 @@ def inverse_kinematics(self, frame_WCF, start_configuration=None,
...
+        else:
+            # sort values for group configuration
+            joint_state = dict(zip(joint_names, joint_positions))
+            group_joint_names = self.get_configurable_joint_names(group)
+            values = [joint_state[name] for name in group_joint_names]
+            configuration = Configuration(values, self.get_configurable_joint_types(group), group_joint_names)
+
+        return configuration.scaled(self.scale_factor)
+
+    def iter_inverse_kinematics(self,
```
romanarust, line 1071: This function looks very specific to your problem.

src/compas_fab/robots/robot.py
```diff
@@ -453,21 +403,53 @@ def get_configurable_joint_types(self, group=None):
         configurable_joints = self.get_configurable_joints(group)
         return [j.type for j in configurable_joints]
 
+    def get_links_distance(self, configuration, link_name_1, link_name_2, group=None):
```
romanarust, line 406: similar here, this is a convenience function, of which I don't see the point of adding to the library.

src/compas_fab/robots/robot.py
gonzalocasas: I was also confused about this at first, but I was chatting with @matteo-pacher and I think it actually makes a lot of sense. The naming of these functions is still to be discussed, in few words, the goal of this was to build a constraint that uses a sphere to achieve an arm position that is fully stretched (ie close to singularity). I think it's a very clever approach to  tackle the problem.

@matteo-pacher perhaps you could expand this explanation on the PR description (maybe adding a few images) to show how this is supposed to work.

romanarust:
fixed the sphere scaling in https://github.com/compas-dev/compas/commit/9ec96151627ee2148378f3986309cf1ca525b246

src/compas_fab/robots/robot.py
```diff
@@ -1014,6 +1068,114 @@ def inverse_kinematics(self, frame_WCF, start_configuration=None,
 
         return configuration.scaled(self.scale_factor)
 
+    def iter_inverse_kinematics(self,
```
gonzalocasas, line 1071: Here, one issue is that `iter_*` functions should return iterators, not the blocking behavior implemented here.

gonzalocasas: This function, as it stands right now, it's not composable because it calls `position_constraint_from_max_reach` internally, which, as @romanarust points out, it's specific to your problem. But if you shuffle things around a little bit, and it will become much more usable.

A quick draft of what I would have expected of this function is this:
```python
def iter_inverse_kinematics(self, ....., max_iterations):
   for i in range(max_iterations):
      yield self.inverse_kinematics(...)
```

This would call ik for a max of the `max_iterations` parameter, but can also be used as a proper iterator, but breaking at the first acceptable result from the calling side will stop the calls from occurring (instead of having to wait for all the iterations to finish). That puts the calling code in control.

Obviously calling repeatedly IK only makes sense in some cases (redundant robot, or random seed).

src/compas_fab/robots/robot.py
```diff
@@ -1014,6 +1068,114 @@ def inverse_kinematics(self, frame_WCF, start_configuration=None,
...
+                configuration = self.inverse_kinematics(frame_WCF,
+                                                        start_configuration=start_configuration,
+                                                        group=group,
+                                                        avoid_collisions=avoid_collisions,
+                                                        constraints=constraints,
+                                                        attempts=attempts,
+                                                        attached_collision_meshes=attached_collision_meshes,
+                                                        full_joint_state=full_joint_state)
+                result_configurations.append(configuration)
+            except BaseException:
```
gonzalocasas, line 1170: This should never be done, catching and passing `BaseException` will hide a lot of problems that should not be hidden.

PR #137, opened 2020-04-13
pydocstyle ignore D100 ”Missing docstring in public module”

tetov:
I'd like to ignore error D105, "Missing docstring in magic method" (i.e. `__repr__`, `__eq__` etc). There is not much to say about these functions, they are what they are. If this is rejected I'd suggest we define some standard docstrings for these methods.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -15,6 +15,7 @@ exclude =
 
 [pydocstyle]
 convention = numpy
+add-ignore = D105
 
 [tool:pytest]
 testpaths = src tests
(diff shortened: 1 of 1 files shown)


gonzalocasas:
I know it might not add a lot of info to document these, but on the other hand, do we have some many that it is a painful process to "just do it"?

tetov:
Haha! Fair point. 

Ok, what’s your opinion on D100, ”Missing docstring in public module”? Usually this info is found in `__init__.py`. Could add something short to all modules if you’d like.

PR #138, opened 2020-04-14
robots/constraints docs

tetov:

- [x] Doc update,

### Checklist

## Notes

I'm [hosting the sphinx docs generated from these file changes](https://p.tetov.se/robots_constraints/reference/compas_fab.robots.html) to make reviewing easier.

I took the liberty to rename `type` in the Constraint classes `__init__` to make it clear that it's not the built-in used. Did not change the attribute name though.


src/compas_fab/robots/constraints.py
```diff
@@ -183,37 +239,56 @@ def __init__(self, joint_name, value, tolerance_above=0., tolerance_below=0., we
...
-        order of [w, x, y, z].
-    tolerances: list of float, optional
-        Error tolerances ti for each of the frame's axes. The respective
-        bound to be achieved is [ai - ti, ai + ti]. Defaults to [0.01, 0.01, 0.01].
-    weight: float, optional
+        order of ``[w, x, y, z]``.
+    tolerances : :class:`list` of :class:`float`, optional
+        Error tolerances t\ :sub:`i` for each of the frame's axes. If only one
+        value is passed it will be used for all 3 axes. The respective bound to
+        be achieved is ``[ai - ti, ai + ti]``. Defaults to
```
beverlylytle, line 281: what about 
> The respective bound to be achieved is :math:`(a_{i} - t_{i}, a_{i} + t_{i})`
or  even ":math:`a_{i} \pm t_{i}`" (assuming standard latex works here)?

src/compas_fab/robots/constraints.py
```diff
@@ -155,24 +211,24 @@ class JointConstraint(Constraint):
...
     ----------
-    joint_name: string
+    joint_name : :class:`str`
         The name of the joint this contraint refers to.
-    value: float
+    value : :class:`float`
         The targeted value for that joint.
-    tolerance_above: float
+    tolerance_above : :class:`float`
         Tolerance above the targeted joint value, in radians. Defaults to 0.
```
beverlylytle, line 219: I think the 0 here and a couple lines below should be surrounded by ticks.

src/compas_fab/robots/constraints.py
tetov: Thanks for reviewing!

Maybe option 1 would be best, to make it easier to read when looking at the source code. What do you think?

beverlylytle: sounds good

src/compas_fab/robots/constraints.py
tetov: c43fb9b

src/compas_fab/robots/constraints.py
tetov: c43fb9b

tetov:
Build fails on [Windows Python 2.7](https://travis-ci.com/github/tetov/compas_fab/jobs/320507740#L316). It fails running `$ pip3 install --no-cache-dir -r requirements-dev.txt || pip install --no-cache-dir -r requirements-dev.txt`, while building [llvm-lite](https://travis-ci.com/github/tetov/compas_fab/jobs/320507740#L316).

Last build it downloaded and installed a wheel, but this time it couldn't find one.

tetov:
llvmlite seems to have [dropped python 2 support](https://github.com/numba/llvmlite/pull/551). It is pulled in by numba. I think we need to pin numba in compas to fix this.

Ping @gonzalocasas and @brgcode.

src/compas_fab/robots/constraints.py
```diff
@@ -12,32 +12,64 @@
...
-        if type not in [self.BOX, self.SPHERE, self.MESH]:
-            raise ValueError("Type must be %d, %d or %d"
-                             % (self.BOX, self.SPHERE, self.MESH))
-        self.type = type
+    #: List of supported volume types
+    VOLUME_TYPES = (BOX, SPHERE, MESH)
+
+    def __init__(self, volume_type, volume):
+        if volume_type not in self.VOLUME_TYPES:
+            raise ValueError("Type must be %d, %d or %d" % (self.VOLUME_TYPES))
```
gonzalocasas, line 57: Maybe best to just print out all, otherwise, adding a new volume type would break it:
```suggestion
            raise ValueError("Type must be one of {}".format(self.VOLUME_TYPES))
```

beverlylytle (approved):

PR #139, opened 2020-04-15
robots/configuration docs

tetov:

- [x] Doc update,

### Checklist

## Notes

I'm [hosting the sphinx docs generated from these file changes](https://p.tetov.se/robots_configuration/reference/compas_fab.robots.html) to make reviewing easier.


src/compas_fab/robots/configuration.py
```diff
@@ -11,16 +11,25 @@
...
+        Joint values expressed in radians or meters, depending on the respective
+        type.
+    types : :obj:`list` of :attr:`compas.robots.Joint.SUPPORTED_TYPES`
+        Joint types, e.g. a list of :attr:`compas.robots.Joint.REVOLUTE` for
+        revolute joints.
     joint_names : :obj:`list` of :obj:`str`, optional
-        Joint names list
+        List of joint names.
+
+    Attributes
```
gonzalocasas, line 28: But now the output is incorrect, because the attributes listed are only `data`,  `prismatic_values` and `revolute_values`, and it's missing the ones set above that you changed from attributes to parameters. The thing about params vs attributes is annoying, but I think it really means we need to duplicate a large chunk of text if we want both documented.

tetov: Sorry I'm not sure I'm following. They are all there but separated? https://p.tetov.se/robots_configuration/reference/generated/compas_fab.robots.Configuration.html#compas_fab.robots.Configuration

Do you want me to copy their docstrings up to class docstring, or am I understanding you wrong?

gonzalocasas: Look at the current output:
image

There are 3 parameters listed, that's correct.
And then there are only 3 attributes (`data`, `prismatic_values` and `revolute_values`) which have no heading. If the attributes would have a heading, then they would be clearly incorrect, because they would be missing the other 3 attributes! (`values`, `types` and `joint_names`)

> Do you want me to copy their docstrings up to class docstring

That is exactly the recommended way according to [this document](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html), which -as I mentioned above- is annoying because it means duplicated a lot of strings, but seems to be "the right way™":

image

tetov: My working assumption was this; if attribute is in `__init__()` add under Parameters, else put under attributes.

Would you like this instead: All attributes goes under Attributes, if they are in `__init__()` they also go under parameters?

gonzalocasas: As the old internet wisdom goes: "assumption is the mother of all fuckups" 😆 

> if attribute is in __init__() add under Parameters, else put under attributes.

`Parameters` and `Attributes` are two different things. When `Parameters` are documenting the `__init__` of a class, there will obviously a large overlap between `Parameters` and `Attributes`, but they still need to be duplicated because they are different things in the end.

tetov: Haha, yeah, good thing this fuckup is only about docstyle conventions then ;).

In my defense, I understand the difference but thought it would be implied that the parameters were also attributes. Looking at the parameters as `__init__()` docstring makes the duplication understandable now to me :).

tetov: ff4bf0b

gonzalocasas: 👍  looking good! is this now ready from your side? or still working on it?

tetov: All done!

tetov:
Done!

beverlylytle (approved):

PR #141, opened 2020-04-15
robots/trajectory docs

tetov:

- [x] Doc update,

### Checklist

## Notes

I'm [hosting the sphinx docs generated from these file changes](https://p.tetov.se/robots_trajectory/reference/compas_fab.robots.html) to make reviewing easier.


src/compas_fab/robots/trajectory.py
```diff
@@ -21,17 +22,19 @@ class JointTrajectoryPoint(Configuration):
...
 
-    Attributes
+    Parameters
     ----------
-    values : :obj:`list` of :obj:`float`
-        Joint values expressed in radians or meters, depending on the respective type.
-    types : :obj:`list` of :class:`compas.robots.Joint.TYPE`
-        Joint types, e.g. a list of :class:`compas.robots.Joint.REVOLUTE` for revolute joints.
-    velocities : :obj:`list` of :obj:`float`
+    values : :class:`list` of :class:`float`
```
gonzalocasas, line 27: What's the difference between `:class:` and `:obj`? I was under the impression the built-in objects are supposed to be documented with `:obj:`, but I might be wrong.

src/compas_fab/robots/trajectory.py
```diff
@@ -45,6 +48,7 @@ def __init__(self, values=None, types=None, velocities=None, accelerations=None,
         self.time_from_start = time_from_start or Duration(0, 0)
 
     def __str__(self):
+        """Return a nicely printable representation of :class:`JointTrajectoryPoint`."""
```
gonzalocasas, line 51:
```suggestion
        """Return a human-readable string representation of the instance."""
```

src/compas_fab/robots/trajectory.py
```diff
@@ -21,17 +22,19 @@ class JointTrajectoryPoint(Configuration):
     Trajectory points are defined either as *values + velocities and
     accelerations*, or as *values + efforts*.
 
-    Attributes
+    Parameters
```
gonzalocasas, line 25: What is the criteria for choosing `Parameters` over `Attributes`? According to the [this doc](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html) it seems to be recommending having BOTH! 😱

tetov: [Numpydocs](https://numpydoc.readthedocs.io/en/latest/format.html#id10) agrees..

Maybe we can think of Parameters as documentation for __init__ only, and then document the attributes for the class (including things straight from __init__)?

src/compas_fab/robots/trajectory.py
tetov: From [Sphinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#cross-referencing-python-objects):

> ###  :py:class:
> Reference a class; a dotted name may be used.
>
> ### :py:obj:
> Reference an object of unspecified type. Useful e.g. as the [default_role](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-default_role).
>
> _New in version 0.4._

My reasoning was that `:class:` is more specific. But then again, are the builtins classes? I'm very open to use `obj` here. I actually had to start googling the difference between class, object and instance because of these.. (As I understand it both an instance and a class is an object..?)

tetov:
Thank you for your comments @gonzalocasas! I think I'll need to put together a section for CONTRIBUTING.md about documentation collecting the conclusions of our discussion in the doc PRs.

src/compas_fab/robots/trajectory.py
tetov: After some more research, yeah let's go with `:obj:`, it seems to be the standard.

gonzalocasas: 👍

src/compas_fab/robots/trajectory.py
gonzalocasas: The "canonical" example for how to use and which style to use is still this document: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

> Maybe we can think of Parameters as documentation for init only, and then document the attributes for the class (including things straight from init)?

Yes, that is correct. This is explained in the sphinx document linked above:

    The __init__ method may be documented in either the class level
    docstring, or as a docstring on the __init__ method itself.

    Either form is acceptable, but the two should not be mixed. Choose one
    convention to document the __init__ method and be consistent with it.

We always use the former option of documenting it at the class level.

gonzalocasas:
> Thank you for your comments @gonzalocasas! I think I'll need to put together a section for CONTRIBUTING.md about documentation collecting the conclusions of our discussion in the doc PRs.

Correct me if I'm wrong, but all we need to do is link to this: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html
I don't think we're deviating from those guidelines (or at least, we aim at not deviating from them ;))

src/compas_fab/robots/trajectory.py
tetov: 34f6156

tetov:
It seems like I did 34f6156 a bit too quick. Fixed now.

tetov:
Done!

tetov:
Vrep doctest fails. 

https://travis-ci.com/github/compas-dev/compas_fab/jobs/345502053#L559

beverlylytle (approved):

PR #143, opened 2020-04-15
Fixes for class docs

tetov:
# Attribute section rendering

While working through docs I noticed that the lists of Attributes in the generated Sphinx documentation looked different compared to class documentation for Compas.

The difference comes from some patching of napoeleon (Sphinx extension) in `docs/conf.py` that @brgcode did. I copied this to compas_fab's `docs/conf.py`.

## Before

before

## After

after

# `__init__()` in class documentation

I also removed the option `'special-members': '__init__'` from  `autodoc_default_options`. The signature is already displayed at the top of the document. Having this option turned on results in the small init section at the bottom, and warnings like this while building the docs:
```
/[...]/compas_fab/src/compas_fab/robots/constraints.py:docstring of
compas_fab.robots.OrientationConstraint.__init__:1:
WARNING: duplicate object description of compas_fab.robots.OrientationConstraint.__init__,
other instance in reference/generated/compas_fab.robots.OrientationConstraint,
use :noindex: for one of them
```
## Removed `__init__()` section
init

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -4,6 +4,8 @@ from __future__ import unicode_literals
 import os
 import sphinx_compas_theme
 
+from sphinx.ext.napoleon.docstring import NumpyDocstring
+
 extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.autosummary',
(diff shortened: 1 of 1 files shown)


gonzalocasas:
yay! 🎉

Issue #142, opened 2020-04-15
Scale method using compas.geometry.Scale or scale factor


tetov:
Classes found in `compas_fab.robots` often have a `scale` method. When it comes to joints, links and constraints of those the scaling is only a multiplication. When it comes to geometry, as in `BoundingVolume` and `CollisionMesh` (examples below), there's two different methods. 

`BoundingVolume`'s method is a bit limiting since it only allows for uniform scaling, but that could be changed by allowing a list of scale factors.

`CollisionMesh` accepts any `compas.geometry.Transformation`. But maybe the `scale` method should only scale the geometry. We can always add a `transform` method, [like in `BoundingVolume`](https://github.com/compas-dev/compas_fab/blob/788739f2d1f34f8975c848f37514df4aa5270334/src/compas_fab/robots/constraints.py#L97).

from [`compas_fab.robots.BoundingVolume`](https://github.com/compas-dev/compas_fab/blob/788739f2d1f34f8975c848f37514df4aa5270334/src/compas_fab/robots/constraints.py#L93)
```python
 def scale(self, scale_factor):
        S = Scale([scale_factor] * 3)
        self.transform(S)
```
from [`compas_fab.robots.CollisionMesh`](https://github.com/compas-dev/compas_fab/blob/788739f2d1f34f8975c848f37514df4aa5270334/src/compas_fab/robots/planning_scene.py#L46)
```python
def scale(self, transformation):
        """Scales the collision mesh.
        Parameters
        ----------
        transformation : compas.geometry.Scale
        """
        mesh_transform(self.mesh, transformation)
```

I can open a PR with these changes if this makes sense :).

romanarust:
Thanks for pointing out the inconsistency!
In that context, `scale` is always used uniformely and actually only used to convert robot units from m to cm, or mm, depending on the preferred planning units. So, scaling by only one factor is sufficient. If you could change the method `CollisionMesh.scale` to take only a scale_factor, that would be great.

But maybe it would be good here to reconsider the `scale` at all:
* should't `scale` be rather `scale_by_factor`?
* or even `scale_to_mm`, `scale_to_cm`, `scale_to_m`?
@gonzalocasas: any thoughts?

tetov:
Fixed in #144

matteo-pacher:
> But maybe it would be good here to reconsider the `scale` at all:
> 
> * should't `scale` be rather `scale_by_factor`?
[...]

PR #144, opened 2020-04-16
Use scale factor for CollisionMesh (suggestion for #142)

tetov:
Suggestion for #142 

Change in a **backwards-compatible** manner.

Not if users are scaling the collision mesh before adding it to scene.

### Checklist

- ~~I have added tests that prove my fix is effective or that my feature works.~~ (Will do that when I write tests for whole module)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -33,6 +33,8 @@ Unreleased
   ``tolerance_below`` for allowing asymmetrical constraints.
 * In :class:``compas_fab.robots.Robot``, changed the ``constraints_from_configuration``
   function with ``tolerances_above`` and ``tolerances_below``.
+* :meth:`compas_fab.robots.CollisionMesh.scale` now takes a scale factor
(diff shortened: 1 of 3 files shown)


tetov:
To make it completely backwards compatible we could do something like this:
```python
class CollisionMesh(object):
    """[...]"""
        
    def scale(self, scale_factor):
        """Scales the collision mesh uniformly.

        Parameters
        ----------
        scale_factor : :obj:`float`
            Scale factor. A :class:`compas.geometry.Scale` instance is also accepted for backwords compatibility.
        """
        if ifinstance(scale_factor, Scale):
            # Scale instance accepted to ensure backwards compatibility
            S = scale_factor
        else:
            S = Scale([scale_factor] * 3)

        self.mesh.transform(S)
```

romanarust:
i would go for a breaking change though....

tetov:
Added line to changelog to make it ready for merge :).

romanarust:
Approved! Thanks!

PR #145, opened 2020-04-16
robots/planning_scene docs

tetov:

- [x] Doc update,

### Checklist

## Notes

I'm [hosting the sphinx docs generated from these file changes](https://p.tetov.se/robots_planning_scene/reference/compas_fab.robots.html) to make reviewing easier.


src/compas_fab/robots/planning_scene.py
```diff
@@ -167,21 +177,24 @@ def remove_collision_mesh(self, id):
...
+        scale : :obj:`bool`, optional
+            If ``True``, the mesh will be scaled according to the robot's scale
             factor.
 
         Returns
         -------
-        None
+        ``None``
+
+        Note
```
gonzalocasas, line 194: I would move this note to be  after the first line of the summary, and would rephrase it to explain more in detail how `append` differs from `add`

tetov: Ok! Reason I keep Notes so far down is because that's the order given in [numpydoc spec](https://numpydoc.readthedocs.io/en/latest/format.html) but I can move them up.

gonzalocasas: I mean, move the text to become part of a multi-line summary text removing the `Note` heading, not that you move the `Note` heading up ;)

tetov: Please check 9a6dff2. I rephrased both add and append to try to make it clearer.

src/compas_fab/robots/planning_scene.py
```diff
@@ -179,6 +211,11 @@ def remove_collision_mesh(self, id):
     def append_collision_mesh(self, collision_mesh, scale=False):
         """Append a collision mesh to the planning scene.
 
+        Adds or appends :class:`CollisionMesh` to :class:`PlanningScene` using
+        `id`. This means that if there's already a :class:`CollisionMesh` in the
+        planning scene it will be merged with the :class:`CollisionMesh` given
+        as an argument to the method.
```
gonzalocasas, line 217:
```suggestion
        Appends a :class:`CollisionMesh` to the :class:`PlanningScene` using
        ``id`` as an identifier of a group or cluster of collision meshes. If the group
        does not exist, it will be created implicitly; if it does exist, the meshes will be
        appended to it instead.
        
        Grouping meshes under a common identifier allows to remove them all
        in one operation, using the :meth:`~PlanningScene.remove_collision_mesh` with 
        the group identifier.
```

src/compas_fab/robots/planning_scene.py
```diff
@@ -92,10 +120,17 @@ def __init__(self, collision_mesh, link_name, touch_links=None, weight=1.):
...
+    Parameters
+    ----------
+    robot : :class:`Robot`
+        A reference to the robot in the planning scene.
+
     Attributes
     ----------
     robot : :class:`Robot`
         A reference to the robot in the planning scene.
+    client : :class:`compas_fab.backend.RosClient` or :class:`compas_fab.backend.VrepClient`
```
gonzalocasas, line 132: Don't refer to specific implementations here. It doesn't add much value and it would mean there's more effort in the event of adding a new client.

src/compas_fab/robots/planning_scene.py
```diff
@@ -59,6 +74,19 @@ def scale(self, transformation):
 class AttachedCollisionMesh(object):
     """Represents a collision mesh that is attached to a :class:`Robot`'s :class:`~compas.robots.Link`.
 
+    Parameters
+    ----------
+    collision_mesh : :class:`compas_fab.robots.CollisionMesh`
+        The collision mesh we want to attach.
```
gonzalocasas, line 80: Try to avoid first persons.

```suggestion
        The collision mesh to be attached to the robot model.
```

src/compas_fab/robots/planning_scene.py
```diff
@@ -59,6 +74,19 @@ def scale(self, transformation):
...
+    link_name : :obj:`str`
+        The name of the :class:`~compas.robots.Link` the collision mesh will be
+        attached to.
+    touch_links : :obj:`list` of :obj:`str`
+        The list of link names the collision mesh is allowed to touch. Defaults
+        to the link it is attached to.
+    weight : :obj:`float`
+        The weight of the attached object. Defaults to ``1.0``.
```
gonzalocasas, line 88: What do these weight do? I think it's unclear from a user side if you read this.

src/compas_fab/robots/planning_scene.py
tetov: From a documentation standpoint having a parent class for backend clients would make sense here. Seems like that might be part of #149. Until that's implemented we can just skip type hint.

src/compas_fab/robots/planning_scene.py
tetov: It comes from weight as defined here: https://docs.ros.org/kinetic/api/moveit_msgs/html/msg/AttachedCollisionObject.html right?

Would it help to add that it's in kg? (Assumption based on [REP 103](https://www.ros.org/reps/rep-0103.html#units)

src/compas_fab/robots/planning_scene.py
gonzalocasas: yep

src/compas_fab/robots/planning_scene.py
gonzalocasas: ah! yes, i think it would. I was initially wondering if this would be a weight as in "weighted-average"

tetov: That was my initial thought as well, but your question prompted me to dig deeper :).

tetov:
Vrep doctest in docs/examples fails. 

https://travis-ci.com/github/compas-dev/compas_fab/jobs/345501934#L558

beverlylytle (approved):

PR #147, opened 2020-04-17
Py27 numba travis fix

tetov:

- [x] Temporary fix of travis test until https://github.com/compas-dev/compas/issues/510 fix is in compas version supported by compas_fab.


tetov:
Closing since this will be solved in compas (see https://github.com/compas-dev/compas/pull/518)

PR #149, opened 2020-04-23
Backend features

beverlylytle:

### Checklist

- [ ] I ran lint on my computer and there are no errors (i.e. `invoke lint`) .


beverlylytle:
@gonzalocasas here is a draft of the backend refactor (clearly not finished yet).  Which bits and pieces do you like?  And which are headed in the wrong direction?

beverlylytle:
To give some context to the ClientManager, it could be used something like 
``` 
with ClientManger(['ros', 'vrep']) as manager:
    robot = manager.ros.load_robot()
    manager.vrep.is_connected
```

src/compas_fab/backends/ros/planner_backend_moveit.py
```diff
@@ -23,6 +23,11 @@ class MoveItPlanner(PlannerInterface):
 
     def __init__(self, client):
         super(MoveItPlanner, self).__init__(client)
+        self.inverse_kinematics = MoveItInverseKinematics(self.client)
+        self.forward_kinematics = MoveItForwardKinematics(self.client)
+        self.plan_cartesian_motion = MoveItPlanCartesianMotion(self.client)
+        self.plan_motion = MoveItPlanMotion(self.client)
+        self.get_planning_scene = MoveItPlanningScene(self.client)
```
gonzalocasas, line 30: looking good!

romanarust:
> To give some context to the ClientManager, it could be used something like
> 
> ```
> with ClientManger(['ros', 'vrep']) as manager:
>     robot = manager.ros.load_robot()
>     manager.vrep.is_connected
> ```

Maybe we should also consider the other way around. E.g. we could load a `compas_fab.Robot` from a client, but in some cases, it could also be beneficial to create a client from a `compas_fab.Robot`. (Especially if we work with multiple clients at the same time, or if we need to retrieve information from the robot)
```
with ClientManager(['pybullet']).from_robot(robot) as manager:
    manager.pybullet.create_collision_map()
```

src/compas_fab/backends/client.py
```diff
@@ -0,0 +1,117 @@
...
+#
+#     def get_configuration(self, *args, **kwargs):
+#         raise NotImplementedError('Assigned control does not have this feature.')
+
+
+class PlannerInterface(object):
+    def __init__(self, client):
+        self.client = client
+
+    def init_planner(self):
```
gonzalocasas, line 72: Perhaps, instead of these `init_*` and `dispose_*` methods, we could consider using an `EventEmitter` and then have the planners subscribe to thos init/dispose events instead?

src/compas_fab/backends/client.py
```diff
@@ -0,0 +1,117 @@
...
+        pass
+
+    def dispose_planner(self):
+        pass
+
+    def validate_response(self, response):
```
gonzalocasas, line 78: Not sure if `validate_response` is so general in its current shape, it's more of a MoveIt thing (at least with that signature)

src/compas_fab/backends/client.py
```diff
@@ -0,0 +1,117 @@
...
+        raise Exception('Assigned planner does not have this feature.')
+
+    def plan_cartesian_motion(self, *args, **kwargs):
+        raise Exception('Assigned planner does not have this feature.')
+
+    # ==========================================================================
+    # collision objects and planning scene
+    # ==========================================================================
+
+    def get_planning_scene(self, *args, **kwargs):
```
gonzalocasas, line 101: how about moving the planning scene interface methods one level below, so that `PlannerInterface` inherits from it, something like `PlanningSceneInterface` or `CollidableWorldInterface` or similar? This would allow the implementation of tools that only do collision checking but cannot do the higher level planning stuff.

gonzalocasas:
> ```
> with ClientManager(['pybullet']).from_robot(robot) as manager:
>     manager.pybullet.create_collision_map()
> ```

While the `ClientManager` is a nice construct, I am not entirely sure I understand its application.
From a user's side, what is exactly the benefit of it considering these two snippets?
```
with ClientManger(['ros', 'vrep']) as manager:
    robot = manager.ros.load_robot()
    print(manager.vrep.is_connected)
```
vs
```
with RosClient() as ros:
    with VrepClient() as vrep:
        robot = ros.load_robot()
        print(vrep.is_connected)
```

beverlylytle:
The point of the ClientManager is to gracefully allocated and deallocate resources when you have multiple context managers.  The issue that could arise in your second examples is that the `__enter__` method could fail for `VrepClient` and then the `__exit__` method of `RosClient` wouldn't necessarily be called, as far as I understand.  I also thought it would be a nicer user experience to not have to write so many nested levels.

src/compas_fab/backends/robot_forward_kinematics.py
```diff
@@ -0,0 +1,12 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+
+class RobotForwardKinematics(object):
```
gonzalocasas, line 6: I'm confused here. Where are these classes (`RobotForwardKinematics`, `RobotInverseKinematics`, `RobotPlanCartesianMotion` and `RobotPlanMotion`) used?

beverlylytle: They aren't.  I added them because of consistency, but they won't remain.

src/compas_fab/backends/ros/backend_features/move_it_plan_motion.py
```diff
@@ -0,0 +1,140 @@
...
+from compas_fab.backends.ros.messages import MotionPlanRequest
+from compas_fab.backends.ros.messages import MultiDOFJointState
+from compas_fab.backends.ros.messages import RobotState
+from compas_fab.backends.ros.messages import TrajectoryConstraints
+from compas_fab.backends.ros.planner_backend import ServiceDescription
+from compas_fab.robots import JointTrajectory
+from compas_fab.robots import Configuration
+
+
+class MoveItPlanMotion(object):
```
gonzalocasas, line 23: These ones will inherit from the base ones, like `PlanMotion`, right?

src/compas_fab/backends/ros/client.py
```diff
@@ -98,21 +98,20 @@ class RosClient(Ros):
     def __init__(self, host='localhost', port=9090, is_secure=False, planner_backend='moveit'):
         super(RosClient, self).__init__(host, port, is_secure)
 
-        # Dynamically mixin the planner plugin into this class
         planner_backend_type = PLANNER_BACKENDS[planner_backend]
-        self.__class__ = type('RosClient_' + planner_backend_type.__name__, (planner_backend_type, RosClient), {})
```
gonzalocasas, line 103: yay! finally gone!

src/compas_fab/backends/ros/backend_features/move_it_plan_motion.py
beverlylytle: Right, I was going through and homogenizing one type of widget at a time.  That's the next one.

gonzalocasas:
> The point of the ClientManager is to gracefully allocated and deallocate resources when you have multiple context managers. The issue that could arise in your second examples is that the `__enter__` method could fail for `VrepClient` and then the `__exit__` method of `RosClient` wouldn't necessarily be called, as far as I understand. I also thought it would be a nicer user experience to not have to write so many nested levels.

Definitely agree on the nesting! On the other point, I believe `__exit__` will also be called in that case.

beverlylytle:
I guess you are right about the call to `__exit__`: https://docs.python.org/3/reference/compound_stmts.html#the-with-statement
And while one could write:
`with RosClient() as ros, VrepClient() as vrep:`
I still feel that the ClientManager could be useful.  Say, if one wanted to add a collision mesh, there could be some sort of global method which calls the appropriate method for each client.

gonzalocasas:
> I still feel that the ClientManager could be useful. Say, if one wanted to add a collision mesh, there could be some sort of global method which calls the appropriate method for each client.

*That* is a good argument! I believe this is/will be relevant for @yijiangh 's Pychoreo integration, since there was the need to simultaneously maintain the planning scene in both ROS and Pybullet.

src/compas_fab/backends/client.py
beverlylytle: I don't think I follow.  The `PlannerInterface` is not creating `abstractmethod`'s in the sense of `ABCMeta` which require implementation in any subclass, but rather is providing default behavior for certain functions which may be overwritten.  So, it would seem that tools that only do collision checking are possible as it is now, and an exception would be raised if the user attempted to call, say, `plan_motion`.

src/compas_fab/backends/client.py
beverlylytle: do you mean something like: 
```
class CollisionObjectListener(object):
    def __init__(self, client):
        self.client = client

    def has_topic(self):
        return hasattr(self.client, 'collision_object_topic') and self.client.collision_object_topic

    def initialize(self):
        if self.has_topic():
            return
        self.client.collision_object_topic = Topic(
            self.client,
            '/collision_object',
            'moveit_msgs/CollisionObject',
            queue_size=None)
        self.client.collision_object_topic.advertise()

    def dispose(self):
        if self.has_topic():
            self.client.collision_object_topic.unadvertise()

    def collision_object(self, collision_object, operation=CollisionObject.ADD):
        if not self.has_topic():
            self.initialize()
        collision_object.operation = operation
        self.client.collision_object_topic(collision_object.msg)
```  
This is happens to be very ROS specific and is not so much about initializing a given planner as much as managing a specific pub/sub situation.  Or do you envision each planner having some list of Listeners, each with an intialize/dispose method called on __enter__/__exit__?

src/compas_fab/backends/ros/backend_features/move_it_forward_kinematics.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def __init__(self, ros_client):
+        self.ros_client = ros_client
+
+    def forward_kinematics(self, configuration, group=None, options={}):  # !!! must find all calls to this and adapt  GHX!!!
+        base_link = options['base_link']
+        link_name = options.get('link_name')
+        return self.forward_kinematics_deprecated(base_link, configuration, group, link_name)
+
+    def forward_kinematics_deprecated(self, base_link, configuration, group, ee_link):
```
beverlylytle, line 33: My intention for leaving these `*_deprecated` methods was for ease of transition for any user's existing scripts, that they would only have to change the method name rather than dig through and figure out exactly how to adapt all the signatures.  But then I went and changed the signatures of some of the deprecated methods, eg `robot` is replaced by `base_link`.  So I see a few possibilities here:  1. Leave it as it is, somewhat misleadingly. 2. Squash all the `*_deprecated` methods with their callers, and make the users work to adapt and leave little trace in the code of how things 'used to be'. 3.  Bootstrap a minimal `robot` instance from the actually required data like `base_link` within the new version of the functions and restore the `*_deprecated` method to its original signature.

src/compas_fab/backends/client.py
beverlylytle: Or did you mean to integrate the `EventEmitterMixin` (which I just stubbled across)?

src/compas_fab/backends/ros/backend_features/move_it_forward_kinematics.py
gonzalocasas: The 2nd and 3rd options sound best. I would slightly prefer the 3rd, but I fear it might take too much time for too little benefit.

beverlylytle: For the third option, is there a reason I shouldn't use something like 
```
robot = self.client.load_robot()
```
in, say, `plan_motion` to use as an argument in `plan_motion_deprecated`?

src/compas_fab/backends/client.py
gonzalocasas: Sorry the delay. I meant the `EventEmitterMixin` :)

src/compas_fab/backends/ros/backend_features/move_it_forward_kinematics.py
gonzalocasas: As discussed earlier today, the 3rd option is going to be too much work for very little gain, let's go with 2 and maintain only a clean, new API.

src/compas_fab/backends/client.py
gonzalocasas: Sorry, you are totally right.

src/compas_fab/backends/ros/backend_features/move_it_inverse_kinematics.py
```diff
@@ -0,0 +1,89 @@
...
+from compas_fab.backends.ros.messages import JointState
+from compas_fab.backends.ros.messages import MultiDOFJointState
+from compas_fab.backends.ros.messages import Pose
+from compas_fab.backends.ros.messages import PoseStamped
+from compas_fab.backends.ros.messages import PositionIKRequest
+from compas_fab.backends.ros.messages import RobotState
+from compas_fab.backends.ros.planner_backend import ServiceDescription
+
+
+class MoveItInverseKinematics(object):
```
gonzalocasas: Does this one still need to inherit from `InverseKinematics`?

src/compas_fab/backends/ros/client.py
```diff
@@ -169,121 +168,6 @@ def load_robot(self, load_geometry=False, urdf_param_name='/robot_description',
...
-
-    def append_collision_mesh(self, collision_mesh):
-        raise NotImplementedError('No planner backend assigned')
-
-    def add_attached_collision_mesh(self, attached_collision_mesh):
-        raise NotImplementedError('No planner backend assigned')
-
-    def remove_attached_collision_mesh(self, id):
-        raise NotImplementedError('No planner backend assigned')
-
```
gonzalocasas, line 286: Happiness is a warn backspace key ;)

docs/examples/04_backends_vrep/01_simulation_with_vrep.rst
```diff
@@ -67,9 +67,11 @@ forward kinematics:
         robot_b = rfl.Robot('B')
         client.set_robot_config(robot_a, config_robot_a)
         client.set_robot_config(robot_b, config_robot_b)
+        group_a = robot_a.model.attr['index']
+        group_b = robot_b.model.attr['index']
 
-        frame_a = client.forward_kinematics(robot_a)
-        frame_b = client.forward_kinematics(robot_b)
+        frame_a = client.forward_kinematics(None, None, group=group_a)
```
gonzalocasas, line 73: Maybe we can refresh the `rfl` module to have `Rfl()` as a fixed, only-for-examples, constructor of a minimal fake model, so that this line has less `None`s, and it looks something like 
```python
robot = Rfl()
client.forward_kinematics(rfl, None, group=group_a)
```

It would also be possible to add configuration to be used on the FK call.

src/compas_fab/backends/ros/client.py
```diff
@@ -151,7 +147,7 @@ def load_robot(self, load_geometry=False, urdf_param_name='/robot_description',
             use_local_cache = True
             path_parts = local_cache_directory.strip(os.path.sep).split(os.path.sep)
             path_parts, robot_name = path_parts[:-1], path_parts[-1]
-            local_cache_directory = os.path.sep.join(path_parts)
+            local_cache_directory = os.path.sep.join([os.sep] + path_parts)
```
beverlylytle, line 150: Does this seem like a reasonable fix for the problem I was having with the robot_description cache being added in the wrong directory?

src/compas_fab/backends/ros/client.py
```diff
@@ -145,13 +182,11 @@ def load_robot(self, load_geometry=False, urdf_param_name='/robot_description',
...
 
-        if local_cache_directory is not None:
-            use_local_cache = True
-            path_parts = local_cache_directory.strip(os.path.sep).split(os.path.sep)
-            path_parts, robot_name = path_parts[:-1], path_parts[-1]
-            local_cache_directory = os.path.sep.join(path_parts)
+        cache_info = LocalCacheInfo.from_local_cache_directory(local_cache_directory)
+        use_local_cache = cache_info.use_local_cache
+        robot_name = cache_info.robot_name
+        robot_name = cache_info.local_cache_directory
```
beverlylytle, line 189: That's a funny name for a robot.

src/compas_fab/backends/ros/backend_features/move_it_plan_motion.py
```diff
@@ -0,0 +1,175 @@
...
+                aco = AttachedCollisionObject.from_attached_collision_mesh(acm)
+                start_state.attached_collision_objects.append(aco)
+
+        # convert constraints
+        goal_constraints = [convert_constraints_to_rosmsg(goal_constraints, header)]
+        path_constraints = convert_constraints_to_rosmsg(path_constraints, header)
+
+        if trajectory_constraints is not None:
+            trajectory_constraints = TrajectoryConstraints(constraints=convert_constraints_to_rosmsg(path_constraints, header))
+            # !!! is this supposed to be trajectory_constraints rather than path_constraints?
```
beverlylytle, line 144: I want to highlight line of code.  I directly copy-pasted the (re)definition of `trajectory_constraints`, but it seems a little off to me.

beverlylytle:
> Maybe we should also consider the other way around. E.g. we could load a `compas_fab.Robot` from a client, but in some cases, it could also be beneficial to create a client from a `compas_fab.Robot`. (Especially if we work with multiple clients at the same time, or if we need to retrieve information from the robot)

@romanarust From what I see in the code so far, it seems that one could do the following:
```
with ClientManager(['pybullet']) as client_manager:
    client_manager.pybullet.compas_fab_robot = robot 
```
Is that not equivalent?  Or am I missing something?

src/compas_fab/backends/ros/backend_features/move_it_plan_motion.py
gonzalocasas: I agree with your, this looks like a typo and it should have been `trajectory_constraints` instead.

src/compas_fab/backends/ros/backend_features/move_it_inverse_kinematics.py
```diff
@@ -0,0 +1,120 @@
...
+        kwargs = {}
+        kwargs['base_link'] = options['base_link']
+        kwargs['frame'] = frame_WCF
+        kwargs['group'] = group
+        kwargs['start_configuration'] = start_configuration
+        kwargs['avoid_collisions'] = options.get('avoid_collisions', True)
+        kwargs['constraints'] = options.get('constraints')
+        kwargs['attempts'] = options.get('attempts', 8)
+        # why is attached_collision_meshes unused?
+        # kwargs['attached_collision_meshes'] = options.get('attached_collision_meshes')
```
beverlylytle, line 86: I believe this is the cause of #170.

src/compas_fab/backends/backend_feature_interfaces.py
```diff
@@ -0,0 +1,116 @@
...
+from __future__ import print_function
+
+from abc import ABCMeta
+from abc import abstractmethod
+
+
+class ForwardKinematics(object):
+    __metaclass__ = ABCMeta
+
+    def __call__(self, configuration, group=None, options={}):
```
gonzalocasas, line 12: I think the empty dictionary is problematic here. Because it's the "same" everywhere, so, if any of these methods modify it, it will affect everything else, if should default to `None` and init to `{}` inside.

src/compas_fab/backends/client_manager.py
```diff
@@ -0,0 +1,75 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+
+import compas
+if compas.is_ironpython():
```
gonzalocasas, line 7: This file I find the most problematic. As mentioned below, `contextlib2` is an open question about safe install to ironpython. And while I understand the joy of the fancy resource handling here, I don't think we have a planned usage that justifies it.

I still think that combinining clients into one new client is something that is not done ad-hoc but the end-user, but more of a decision by developers trying to add support for something new, in which case they will code a new class and use that instead.

src/compas_fab/backends/vrep/backend_features/vrep_plan_motion.py
```diff
@@ -0,0 +1,201 @@
...
+                          'sbl', 'stride', 'trrt')
+
+    def __init__(self, client):
+        self.client = client
+
+    def plan_motion(self, goal_constraints, start_configuration=None, group=None, options={}):
+        """Finds a path plan to move the selected robot from its current position to the `goal_constraints`.
+
+        Args:
+            goal_constraints (:class:`Frame`): Target or goal frame.
```
gonzalocasas, line 34: If would be ideal if this could actually be a list of `Constraints` and they check for positional/orientation values to rebuild the frame (and it would allow to merge this and `plan_motion_to_configs` into one method. However, if you decide to skip and leave for the future, it would be totally understandable.

src/compas_fab/backends/ros/backend_features/move_it_forward_kinematics.py
```diff
@@ -0,0 +1,85 @@
...
+            The full configuration to calculate the forward kinematic for. If no
+            full configuration is passed, the zero-joint state for the other
+            configurable joints is assumed.
+        group : str, optional
+            The planning group used for the calculation. Defaults to the robot's
+            main planning group.
+        options : dict, optional
+            Dictionary containing the following key-value pairs:
+
+            - backend :: None or str
```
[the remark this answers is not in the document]
beverlylytle: A single colon in some instances has some strange behavior. The following in a doc string results in the attached image: 
```
"""
        options: dict, optional
            Dictionary containing the following key-value pairs:

            - base_link (:obj:`str`) : Name of the base link.
            - avoid_collisions : bool, optional
                Whether or not to avoid collisions. Defaults to `True`.
            - constraints :: list of :class:`compas_fab.robots.Constraint`, optional
                A set of constraints that the request must obey. Defaults to `None`.
"""
```

Screenshot 2020-06-08 at 12 03 53

src/compas_fab/backends/vrep/backend_features/vrep_plan_motion.py
beverlylytle: I believe this is beyond the scope of this PR, and should be made into a separate issue to be addressed in the future.

gonzalocasas: agree 👍

src/compas_fab/backends/ros/backend_features/move_it_forward_kinematics.py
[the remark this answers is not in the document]
gonzalocasas: uuuuhhhggg

src/compas_fab/backends/backend_feature_interfaces.py
```diff
@@ -0,0 +1,116 @@
...
+        return self.forward_kinematics(configuration, group, options)
+
+    @abstractmethod
+    def forward_kinematics(self, configuration, group=None, options={}):
```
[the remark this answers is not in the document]
beverlylytle: Expected return type?  I didn't think there was to be an expected return type.  For example, Vrep's `plan_motion` returns a list of `Configuration`s while for MoveIt, you get a `JointTrajectory`.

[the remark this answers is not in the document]
gonzalocasas: `JointTrajectory` is actually not much more than a class for a list of `Configuration`s, so we could easily update vrep planner to return it (`JointTrajectory` contains points, and each `JointTrajectoryPoint` is a class that inherits from `Configuration`).

But even if that part is out of scope, we could document the return that that we want as defined interface, and then slowly update the non-conformant parts later on.

gonzalocasas:
As discussed, the travis fail is totally unrelated to this pull request, so, from my side I think this is good to go! 

awesome

Issue #159, opened 2020-06-04
1.2.7. Links - link dead


funkchaser:
relink needed:
ABB: What is a coordinate system?

PR #172, opened 2020-06-06
Linkcheck and warnings

tetov:
Fixed links using linkcheck in Sphinx. Lots of http -> https and some real fixes. 

(My philosophy is that using URLs which is redirected at target webpage is just a broken link waiting to happen, so all URLs have been changed to their redirectes.)

Also replaces broken link mentioned in #159.

Doc update

### Checklist

- ~~I added a line to the `CHANGELOG.rst` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).~~
- [ ] ~~I added new functions/classes and made them available on a second-level import, e.g. `compas_fab.robots.Configuration`.~~
- ~~I have added tests that prove my fix is effective or that my feature works.~~
- ~~I have added necessary documentation (if appropriate)~~


tetov:
Need to test if http -> https broke description loading in integration tests.

tetov:
@gonzalocasas: Reasoning behind: 0b0c101

We could use regex instead, something like b"xmlns=http[s]?:\/\/..." but that will be a pain to keep compatible between 2 and 3. 

The schema could be specified with either, the http url redirects to https. Catching both like this is ugly but it also works, haha.

gonzalocasas:
Is this still relevant?

tetov:
> 
> 
> Is this still relevant?

Now it's as relevant as it was when I opened it, so yeah a little bit? Haha!

src/compas_fab/backends/ros/fileserver_loader.py
```diff
@@ -214,6 +214,7 @@ def load_mesh(self, url):
             # XML parsers, so, we just get rid of it for DAE files
             if file_extension == 'dae':
                 file_content = file_content.replace(b'xmlns="http://www.collada.org/2005/11/COLLADASchema"', b'')
+                file_content = file_content.replace(b'xmlns="https://www.collada.org/2005/11/COLLADASchema"', b'')
```
gonzalocasas, line 217: Almost feel the urge to suggest using regex, but no, I will restrain myself :P

gonzalocasas (approved):
James Franco Thumbs Up

tetov:
> @gonzalocasas: Reasoning behind: [0b0c101](https://github.com/compas-dev/compas_fab/commit/0b0c101a5a2d1b0b4d1e06efeb33e46717ddc1fa)
> 
> We could use regex instead, something like b"xmlns=http[s]?://..." but that will be a pain to keep compatible between 2 and 3.
> 
> The schema could be specified with either, the http url redirects to https. Catching both like this is ugly but it also works, haha.

I wrote this, but I don't remember the compat problem. If I know myself I tried regex before `str.replace`.

gonzalocasas:
@tetov  :+1:

PR #181, opened 2020-06-24
Pybullet

beverlylytle:
## What type of change is this?

### Checklist

Here's a version of a basic PyBullet client.  There is some conflict between what's owned and managed by `Robot` and by `PyBulletClient`, so if you have ideas on how to deal with that, I'd be interested.  I've left the documentation a bit sparse until those things are resolved. I also think the way `redirect_stdout` is handled with respect to pytest isn't the most straightforward, so commentary there is particularly welcome.


yijiangh:
@beverlylytle Just curious: what's the main reason for directly interfacing `pybullet` but not through `pybullet_planning`? Trying to figure out what `pybullet_planning` is missing that makes the integration hard.

beverlylytle:
@yijiangh Hi Yijiang!  There are a couple of reasons that went into the decision to using pybullet directly in this attempt at a pybullet client.  First was a desire for a lightweight and basic client in the first iteration, including simplified dependencies--requiring pybullet alone is simpler than requiring pybullet_planning and all of its dependencies, including pybullet.  As you can see from Gonzalo's many commits, we are having a spot of trouble even with the "simple" version of the dependencies.  Secondly, we've just refactored the way we integrate backend clients into compas_fab with the intent of making them as consistent as possible, eg all versions of `add_collision_mesh` should take the same arguments and return the same type.  Things are always evolving, so when we need an animal from your zoo, it's likely we'll integrate pybullet_planning.

yijiangh:
@beverlylytle Gotcha! Based on our experience, `pybullet`'s exposed API is a bit too bare-bone for planning-related research, which is exactly why we spent much time building `pybullet_planning`. For the compas_fab type consistency, I was building a bunch of two-way conversion functions to do that last summer, but I'm sure you guys have more principled ways to do that.

In any event, feel free to send me questions on Slack and I am happy to have a zoom chat any time 😺

beverlylytle:
Another thing I have a question about is now when you run something like `from compas.backends import RosClient` this causes `pybullet` to be imported in the background and so you see `pybullet build time: ***` in the console.  Is that something I should try to avoid?

gonzalocasas:
Mmm... The printed output is not important but I guess triggering all the
import machinery for every backend on the first load is not a great idea,
now we have 3 backends but the more we add, the less acceptable it will be.
We need to think how to trigger this as lazily as possible

src/compas_fab/utilities/lazy_loader.py
```diff
@@ -0,0 +1,44 @@
+# Code copied from https://github.com/tensorflow/tensorflow/blob/master/tensorflow/python/util/lazy_loader.py
```
beverlylytle, line 1: This guy, brazenly copied from tensorflow, lazily loads pybullet, but it does not, at least as used, lazily load PyBulletClient when importing anything from compas_fab.backends.

src/compas_fab/backends/pybullet/conftest.py
```diff
@@ -0,0 +1,5 @@
+import compas_fab.backends
```
beverlylytle, line 1: I am wondering if I should pull this conftest up to the top `compas_fab` directory rather than repeating it twice.  I'm conflicted because it's generally a terrible idea to alter behavior because one is running tests rather than in production and I don't want to encourage the use of this thing, but on the other hand, duplicate code is hard to maintain. Or maybe there is a different solution entirely?

gonzalocasas: In the case of conftest files, I think it's fair to duplicate.

src/compas_fab/backends/pybullet/client.py
```diff
@@ -0,0 +1,410 @@
...
+        temp_dir = tempfile.mkdtemp()
+        tmp_obj_path = os.path.join(temp_dir, 'temp.obj')
+        mesh.to_obj(tmp_obj_path)
+        pyb_body_id = self.body_from_obj(tmp_obj_path)
+        self.set_base_frame(frame, pyb_body_id)
+        # !!! The following lines are apparently for visual debugging purposes
+        # To be deleted or rewritten later.
+        # if name:
+        #     pybullet_planning.add_body_name(pyb_body, name)
+        os.remove(tmp_obj_path)
```
gonzalocasas, line 349: wrap this temp dir stuff into a `try..finally` block so that we make sure we remove the temp dir/file even if the middle part crashes.

src/compas_fab/backends/pybullet/client.py
```diff
@@ -0,0 +1,410 @@
...
+        body_id = self.body_id_or_default(body_id)
+        point, quaternion = pose_from_frame(frame)
+        pybullet.resetBasePositionAndOrientation(body_id, point, quaternion, physicsClientId=self.client_id)
+
+    def set_joint_position(self, joint_id, value, body_id=None):
+        body_id = self.body_id_or_default(body_id)
+        pybullet.resetJointState(body_id, joint_id, value, targetVelocity=0, physicsClientId=self.client_id)
+        # self.create_collision_map() ???
+
+    def set_joint_positions(self, joints, values, body_id=None):
```
gonzalocasas, line 318: Please add docstrings to all of these ones if they are intended to be public, otherwise, prefix with `_`.

src/compas_fab/backends/pybullet/client.py
```diff
@@ -0,0 +1,410 @@
...
+    @property
+    def is_connected(self):
+        """Indicates whether the client has an active connection.
+
+        Returns
+        -------
+        :obj:`bool`
+            ``True`` if connected, ``False`` otherwise.
+        """
+        return self.check_connection() == 1
```
gonzalocasas, line 95: This fails if it's never connected, because `client_id` is `None`:
```
>>> client = PyBulletClient()
>>> client.is_connected
pybullet build time: Jul  8 2020 13:56:47
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "c:\users\gcasas\eth\projects\research\compas_fab\src\compas_fab\backends\pybullet\client.py", line 95, in is_connected
    return self.check_connection() == 1
  File "c:\users\gcasas\eth\projects\research\compas_fab\src\compas_fab\backends\pybullet\client.py", line 98, in check_connection
    return pybullet.getConnectionInfo(physicsClientId=self.client_id)['isConnected']
TypeError: an integer is required (got type NoneType)
```

src/compas_fab/backends/pybullet/utils.py
```diff
@@ -0,0 +1,76 @@
...
+    ...     print("from Python")                                         # doctest: +SKIP
+    ...     os.system("echo non-Python applications are also supported") # doctest: +SKIP
+    """
+    def _redirect_stdout(to_):
+        sys.stdout.close()  # + implicit flush()
+        os.dup2(to_.fileno(), fd)
+        sys.stdout = os.fdopen(fd, 'w')
+
+    # Pytest interferes with file descriptor capture.
+    if not enabled or compas_fab.backends._called_from_test:
```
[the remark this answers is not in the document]
beverlylytle: I like the brevity, but I can imagine that I will forget what that means in a few weeks time.

src/compas_fab/backends/pybullet/client.py
```diff
@@ -0,0 +1,410 @@
...
+    def load_robot_from_urdf(self, urdf_filename):
+        """Create a pybullet robot using the input urdf file.
+
+        Parameters
+        ----------
+        urdf_filename : :obj:`str`
+            Absolute file path to the urdf file. The mesh file can be linked by either
+            `"package::"` or relative path.
+        """
+        # TODO: fuse end effector and robot link tree into one
```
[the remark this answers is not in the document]
beverlylytle: This was a comment from `pybullet_planning` that I wasn't sure what to do with... Will delete.

src/compas_fab/backends/pybullet/utils.py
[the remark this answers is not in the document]
gonzalocasas: how about this then? It's more or less the same as you have currently, but does not depend on shared global state ;)

```python
    called_from_test = 'pytest' in sys.modules
    if not enabled or called_from_test:
```

src/compas_fab/backends/pybullet/client.py
```diff
@@ -0,0 +1,410 @@
...
+        options = self.compose_options(color, width, height)
+        with redirect_stdout():
+            self.client_id = pybullet.connect(method, options=options)
+        if 0 > self.client_id:
+            raise Exception('Error in establishing connection with PyBullet.')
+        if self.use_gui:
+            self.configure_debug_visualizer(shadows)
+
+    def detect_display(self):
+        if self.use_gui and system != 'darwin' and not compas.is_windows() and ('DISPLAY' not in os.environ):
```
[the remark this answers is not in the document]
beverlylytle: I had assumed that this was to prevent troubles when running PyBullet on a headless server without openGL support, that is to say, not Travis related. But perhaps we should expect the user to deal with that and delete this check. This is adapted from @yijiangh 's work, so maybe he can explain more.

src/compas_fab/backends/pybullet/client.py
beverlylytle: In testing my fix for this, I noticed the following:  
Screenshot 2020-07-09 at 15 14 57
Should I try to fix this? I fear it might be tricky...

gonzalocasas: I'm not sure I understand what you are pointing out on your comment. The output looks alright.. doesn't it? (except the fact that the printouts are all after the `.disconnect()` instead of after each line but I assume that is totally unrelated)

beverlylytle: Yeah, it's that the output comes after `disconnect`.  I'm sure it has something to do with `redirect_stdout` and is unrelated to `is_connected`, but this is just how I noticed it.

gonzalocasas: ah yes, it is probably related to buffering flags of `stdout` (there's a comment about it in the `finally` block of `redirect_stdout`). Maybe you can try to disable buffering in line 36 with something like `sys.stdout = os.fdopen(fd, 'w', buffering=0)`.

src/compas_fab/backends/pybullet/client.py
```diff
@@ -0,0 +1,410 @@
...
+            for link2_name in names:
+                link2 = self.link_id_by_name[link2_name]
+                pts = pybullet.getClosestPoints(bodyA=self.robot_uid, bodyB=self.robot_uid, distance=0,
+                                                linkIndexA=link1, linkIndexB=link2, physicsClientId=self.client_id)
+                if pts:
+                    LOG.warning("Collision between '{}' and '{}'".format(link1_name, link2_name))
+                    return True, (link1_name, link2_name)
+
+        # attached collision objects
+        # !!! is this right?
```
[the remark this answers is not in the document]
beverlylytle: it wasn't ;)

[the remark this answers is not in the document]
gonzalocasas: ;)

src/compas_fab/robots/semantics.py
```diff
@@ -128,6 +129,12 @@ def __get_passive_joints(self):
     def __get_end_effectors(self):
         return [ee.attrib['parent_link'] for ee in self.root.findall('end_effector')]
 
+    def __get_disabled_collisions(self):
+        return {
+            frozenset([dc.attrib['link1'], dc.attrib['link2']])
```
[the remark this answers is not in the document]
beverlylytle: I don't think I understand your understanding ;)  `link1` and `link2` are the argument names used within srdf's `disable_collision` tag.

src/compas_fab/backends/pybullet/client.py
beverlylytle: `ValueError: can't have unbuffered text I/O`

src/compas_fab/robots/semantics.py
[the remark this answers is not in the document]
gonzalocasas: yep, got it. the names stand for `first-link-in-pair` and `second-link-in-pair`.   I initially confused `link1` as meaning `link1 to linkn` in the robot model.

src/compas_fab/backends/pybullet/client.py
gonzalocasas: 😬

src/compas_fab/backends/pybullet/client.py
```diff
@@ -0,0 +1,509 @@
...
+    --------
+    >>> from compas_fab.backends import PyBulletClient
+    >>> with PyBulletClient(connection_type='direct') as client:
+    ...     print('Connected: %s' % client.is_connected)
+    Connected: True
+
+    """
+    def __init__(self, connection_type='gui', verbose=False):
+        super(PyBulletClient, self).__init__(connection_type)
+        self.planner = PyBulletPlanner(self)
```
gonzalocasas, line 139: Eventually (not in this PR), this could be the point in which we hook up planners as plugins: the pybullet client could define a `@pluggable` like `create_pybullet_planner()` and by default a built-in plugin would provide it `PyBulletPlanner` instances, but if @yijiangh 's upcoming `compas_fab_pychoreo` is installed and provides another implementation, it could potentially be selected over the default one.

src/compas_fab/backends/pybullet/client.py
```diff
@@ -0,0 +1,509 @@
...
+
+    def _add_ids_to_robot_links(self, robot):
+        joint_ids = self._get_joint_ids(robot.pybullet_uid)
+        for link_id in joint_ids:
+            link_name = self._get_link_name(link_id, robot.pybullet_uid)
+            link = robot.model.get_link_by_name(link_name)
+            pybullet_attr = {'id': link_id}
+            link.attr.setdefault('pybullet', {}).update(pybullet_attr)
+
+    def remove_configurations_in_collision(self, robot, configurations):
```
gonzalocasas, line 204: Minor nitpicking, but I would name this `filter_configurations_in_collision` instead.

src/compas_fab/backends/pybullet/client.py
```diff
@@ -0,0 +1,509 @@
...
+            # To be deleted or rewritten later.
+            # if name:
+            #     from pybullet_planning import add_body_name
+            #     add_body_name(pyb_body, name)
+        finally:
+            shutil.rmtree(tmp_dir)
+        return pyb_body_id
+
+    @staticmethod
+    def _handle_concavity(tmp_obj_path, tmp_dir, concavity, mass):
```
gonzalocasas, line 413: 👍

Issue #183, opened 2020-07-03
Migrate from travis to Github Actions


gonzalocasas:
We need to migrate the continuous integration build from Travis to Github Actions (mostly triggered by the fact that Travis uses Windows Server Core as their Windows runner, which doesn't support a whole bunch of UI related stuff and failed on the Pybullet integration).

PR #185, opened 2020-07-08
GitHub actions

beverlylytle:
Addresses issue #183.


beverlylytle:
I broke the workflows up into separate files because it seems then you can re-run just the jobs described in one file (say, the job which runs the integration tests) rather than every single job.

tetov:
Wow! Would have thought this would have taken longer! Kudos @beverlylytle!

beverlylytle:
Thanks!

PR #195, opened 2020-09-02
Unify backend feature implementations of FK/IK between Pybullet/ROS

gonzalocasas:
The main goal of this PR is to remove the need to pass `options=dict(robot=robtot)` to IK/FK calls using the PyBullet client and homogenize them so that the dispatching methods of `Robot` class also work with it (so that scaling is taken care of).

I also refactored `RobotSemantics` to make it more like a data class, and less an XML accessor thing.

### Checklist


gonzalocasas:
CI doesn't kick in on this one because it merges to something else?

src/compas_fab/backends/ros/backend_features/move_it_plan_motion.py
```diff
@@ -36,11 +36,13 @@ class MoveItPlanMotion(PlanMotion):
     def __init__(self, ros_client):
         self.ros_client = ros_client
 
-    def plan_motion(self, goal_constraints, start_configuration=None, group=None, options=None):
+    def plan_motion(self, robot, goal_constraints, start_configuration=None, group=None, options=None):
         """Calculates a motion path.
 
         Parameters
         ----------
+        robot : :class:`compas_fab.robots.Robot`
```
beverlylytle, line 44: Now that `robot` is here, can't `base_link`, `joint_names` and `joint_types` all be retrieved from it?  If so, then they wouldn't need to be a part of `options`.  I think such adjustments could be made in many of the ros backend features.

beverlylytle:
I like the refactoring of the semantics.

beverlylytle:
About the CI, yes.  We have it specified in our workflows that it is only triggered on PRs to master.

src/compas_fab/backends/pybullet/backend_features/pybullet_inverse_kinematics.py
```diff
@@ -56,22 +55,20 @@ def inverse_kinematics(self, frame_WCF, start_configuration=None, group=None, op
 
         Returns
         -------
-        :class:`compas_fab.robots.Configuration`
-            The robot's full configuration.
+        tuple of joint positions and joint names
```
beverlylytle, line 58:
```suggestion
        :obj:`tuple` of :obj:`list`
```

src/compas_fab/backends/ros/backend_features/move_it_inverse_kinematics.py
```diff
@@ -69,8 +71,8 @@ def inverse_kinematics(self, frame_WCF, start_configuration=None, group=None, op
 
         Returns
         -------
-        :class:`compas_fab.robots.Configuration`
-            The planning group's configuration.
+        tuple of joint positions and joint names
```
beverlylytle, line 74:
```suggestion
        :obj:`tuple` of :obj:`list`
```

src/compas_fab/robots/robot.py
```diff
@@ -979,24 +979,24 @@ def inverse_kinematics(self, frame_WCF, start_configuration=None, group=None, op
...
         group: str, optional
             The planning group used for calculation. Defaults to the robot's
             main planning group.
+        return_full_configuration : bool, optional
+            If ``True``, returns a full configuration with all joint values
+            specified, including passive ones if available. Defaults to ``False``.
         options: dict, optional
             Dictionary containing the following key-value pairs:
 
             - avoid_collisions :: bool, optional
```
beverlylytle, line 992: Here the documentation isn't in the correct style, but it may have already been fixed on the main branch.

beverlylytle (changes requested):
I like that `robot` has been pulled out of options and added as an argument to many of the backend features.   However in at least the one spot (`move_it_plan_motion`), that robot now exists as an argument makes some of the things that are required in `options` redundant.  Otherwise my comments are small things about documentation.

src/compas_fab/backends/ros/backend_features/move_it_plan_motion.py
beverlylytle: You can see [here](https://github.com/compas-dev/compas_fab/blob/c43fb9b9ade629dd2a84c8936178f80cde008ef0/src/compas_fab/backends/ros/planner_backend_moveit.py) and [here](https://github.com/compas-dev/compas_fab/blob/c43fb9b9ade629dd2a84c8936178f80cde008ef0/src/compas_fab/backends/vrep/client.py) what it was like before the big refactoring when `robot` was still an argument to these functions.

gonzalocasas: Done! `base_link` is still honored if the user decides to pass it as option, but now the `Robot` class needs not to know about it, so, only the MoveIt features, that can actually do something with it check if it is defined, otherwise default to using `robot` to get the base link from the root of the model.

src/compas_fab/backends/ros/backend_features/move_it_plan_motion.py
```diff
@@ -131,9 +137,9 @@ def plan_motion_async(self, callback, errback,
                        goal_constraints=goal_constraints,
                        path_constraints=path_constraints,
                        trajectory_constraints=trajectory_constraints,
-                       planner_id=options.get('planner_id', ''),
+                       planner_id=options.get('planner_id', 'RRTConnectkConfigDefault'),
                        group_name=group,
-                       num_planning_attempts=options.get('num_planning_attempts', 8),
+                       num_planning_attempts=options.get('num_planning_attempts', 1),
```
gonzalocasas, line 142: This was a mismatch to the the documented default (but it worked because it was also defaulted at a higher layer).

src/compas_fab/robots/robot.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, model, artist=None, semantics=None, client=None):
         self.artist = artist
         self.semantics = semantics
         self.client = client
-        self.pybullet_uid = None
+        self.attributes = {}
```
gonzalocasas, line 58: Moved `pybullet_uid` to attributes, so that `Robot` doesn't need to know about it.

src/compas_fab/robots/robot.py
```diff
@@ -1072,28 +1047,51 @@ def inverse_kinematics_deprecated(self, frame_WCF, start_configuration=None,
 
         return configuration.scaled(self.scale_factor)
 
+    def inverse_kinematics_deprecated(self, frame_WCF, start_configuration=None,
```
gonzalocasas, line 1050: I reversed the call hierarchy between `xyz` and their `xyz_deprecated` method, now deprecated methods call the non-deprecated versions, so that we can easily delete them on a future release. It also makes the non-deprecated code simpler because it doesn't need to know much about what's in the `options`.

src/compas_fab/robots/robot.py
```diff
@@ -1072,28 +1047,51 @@ def inverse_kinematics_deprecated(self, frame_WCF, start_configuration=None,
...
             Dictionary containing the following key-value pairs:
 
-            - backend :: None or str
-                If `None` calculates fk with the client if it exists or with the robot model.
-                If 'model' use the robot model to calculate fk. Anything else is open
-                for implementation, possibly 'kdl', 'ikfast'
-            - ee_link :: str, optional
-                The name of the link to calculate the forward kinematics for.
-                Defaults to the group's end effector link.
+            - ``"solver"``: (:obj:`str`, optional) If ``None`` calculates FK
```
gonzalocasas, line 1081: I renamed `backend` to `solver` because one backend can implement multiple solvers, and while the keyword `model` has specific meaning currently, potentially, existing backends can take values of `solver` and select the solving algorithm (the same way the `plan_motion` calls use `planner_id`)

src/compas_fab/sensors/baumer.py
```diff
@@ -433,13 +433,7 @@ def format_command(self, address, command, data=None):
         return self.FRAME_TAIL % (frame, self.calculate_checksum(frame))
 
     def calculate_checksum(self, command):
-        """Checks that message is complete.
-
-        Note
-        ----
-        See `RS-485 Protocol Structure <https://www.baumer.com/ae/en/product-overview/2d-3d-sensors/profile-sensors-/measurement-of-round-objects/\
```
gonzalocasas, line 440: Running `check-links` showed this one as 404, and I couldn't be bother to find it again. 😛

docs/examples/03_backends_ros/06_ros_in_grasshopper.rst
```diff
@@ -0,0 +1,24 @@
+*******************************************************************************
```
gonzalocasas, line 1: I deleted the `full_examples` page because the integration tests were running them against the wrong robot (they were created for the `Panda`, not the `UR5`). Surprisingly, they worked (because we don't check output on those integration tests), but it was very confusing and since the loading of robots (compared to the time in which the first examples were written), has now moved totally to the ros loader -which was partially the goal of the full examples-, I simply moved that to the normal examples and kept only one set of them using UR5. And so I had to put the Grasshopper example in a page of its own.

gonzalocasas:
I think on my side, I'm done for a re-review if you have time. I tried to comment on most of the decisions that might not be super clear.

src/compas_fab/sensors/baumer.py
beverlylytle: i've never looked at this file, but from this snippet, this particular method looks somewhat misleading....

beverlylytle (approved):

src/compas_fab/sensors/baumer.py
gonzalocasas: The sensor supports checksums on the messages, but it also allows `'****'` to skip checking it.

docs/reference.rst
```diff
@@ -4,7 +4,42 @@
 API Reference
 *************
 
+This reference manual details functions, modules, and objects included in COMPAS FAB,
+describing what they are and what they do.
+
+For learning how to use COMPAS FAB, see the :ref:`examples <examples>` and
```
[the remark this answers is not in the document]
gonzalocasas: Fixed on https://github.com/compas-dev/compas_fab/pull/195/commits/5c8115fc9d9dff75e4fe2e664879ff3938b4be52

src/compas_fab/robots/robot.py
```diff
@@ -1048,15 +1026,12 @@ def inverse_kinematics_deprecated(self, frame_WCF, start_configuration=None,
             else:
                 attached_collision_meshes = [self.attached_tool.attached_collision_mesh]
```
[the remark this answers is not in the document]
gonzalocasas: Good point, I applied this change and two other occurrences of the same thing in https://github.com/compas-dev/compas_fab/pull/195/commits/d7dfb8404cfa4e68522cd694a05c96d2b8b78f8f

docs/examples/04_backends_vrep/01_simulation_with_vrep.rst
```diff
@@ -124,16 +124,12 @@ Here is an example of such a request:
 
     with VrepClient() as client:
         robot = rfl.Robot('B')
-        client.set_robot_config(robot, start_config)
         group = robot.model.attr['index']
```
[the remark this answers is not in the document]
gonzalocasas: I fully agree, the V-REP integration needs some refactor to be aligned to the other two. It used to be that (for v-rep planner), `Robot` instance was what MoveIt considers a group, so, the robot model had an "index" which is basically a group name inside a full robot cell/setup.

I'll see if I can bring it into compliance with the rest -a bit at least.  :)

beverlylytle:
I know it's mostly because ghx files are super verbose, but "+1,814 −24,174" must feel good.

PR #196, opened 2020-09-07
Doc versioning

gonzalocasas:
Use the newly added doc versioning option/widget of `sphinx-compas-theme 0.9` to allow switching to different versions of the documentation. I manually added some of the older doc builds to the doc github pages (only to be able to have something to switch on 😛 )

It now looks like this:
image

### Checklist


docs/conf.py
```diff
@@ -68,6 +68,15 @@
     "package_title": project,
     "package_version": release,
     "package_repo": 'https://github.com/compas-dev/compas_fab',
+    "package_docs": 'https://gramaziokohler.github.io/compas_fab',
+    "package_old_versions": [
```
beverlylytle, line 72: Is it possible to put in a hook to update this on each new release?

gonzalocasas:
I fixed the docs github actions workflow to publish versioned docs. Docs from pull requests are published to `https://gramaziokohler.github.io/compas_fab/pull_<###>`, tags are published to the tag name (eg `https://gramaziokohler.github.io/compas_fab/<tag_name>`) and main branch is published to `https://gramaziokohler.github.io/compas_fab/`latest`.

This PR in particular has been published to https://gramaziokohler.github.io/compas_fab/pull_196

docs/doc_versions.txt
```diff
@@ -0,0 +1,6 @@
+latest
+stable
```
beverlylytle, line 2: What does "stable" mean in the context of not yet having a Version 1.0?

docs/conf.py
```diff
@@ -68,6 +76,8 @@
     "package_title": project,
     "package_version": release,
     "package_repo": 'https://github.com/compas-dev/compas_fab',
+    "package_docs": package_docs_root,
+    "package_old_versions": package_docs_versions
```
beverlylytle, line 80: Doesn't "package_old_versions" also contain "latest"?  Is that already old?

docs/doc_versions.txt
gonzalocasas: `stable` is supposed to be the latest *released* version (vs `latest` being bleeding edge from main branch). The idea is to have a permalink into the newest release, but right now, no commit automatically generates it. 

I am not sure if we can use the `tag` event to push at the same time to `stable` and assume that "last tag == stable release"

beverlylytle: Ah, so the intention is that 0.11.0 should be the same thing as "stable"?

docs/conf.py
gonzalocasas: Maybe a better name would have been `package_all_versions` 😛  (this variable name is [defined in the theme](https://github.com/compas-dev/sphinx_compas_theme/blob/master/sphinx_compas_theme/compaspkg/theme.conf))

docs/doc_versions.txt
gonzalocasas: yep! I believe that's the way readthedocs does/names them as well.

beverlylytle (approved):

PR #197, opened 2020-09-07
add scaled to CollisionMesh and use in PlanningScene

### LINKED ISSUE #186: scaling in add_attached_collision_mesh modified the underlying mesh
I encountered a problem recently where repeatedly calling the `add_attached_collision_mesh ` function with the same `mesh` object causes the acm to disappear. And I pinned the cause to this line where the scale function causes an undocumented side effect. 

https://github.com/compas-dev/compas_fab/blob/e4bfe769103aa41ff8beb823864245a4a6d82785/src/compas_fab/robots/planning_scene.py#L231

As a feature suggestion, I think it is more robust if the line uses `scaled`. This allows users to reuse the input mesh object for other purpose such as visualization or in my case not having problem when calling the function repeatedly.

The behavior of this function with the Moveit planner is that the new `acm` replaces the old one if the id is the same. I think it is better to document this. At the moment, it is ambiguous if user should call `RemoveAttachedCollisionMesh` before calling `add_attached_collision_mesh`. From what I have tested (and also @matteo-pacher  told me) it is not necessary.

beverlylytle:
Closes #186 .

### Checklist

### DIFF
--- a/src/compas_fab/robots/planning_scene.py
+++ b/src/compas_fab/robots/planning_scene.py
@@ -70,6 +70,17 @@ class CollisionMesh(object):
         S = Scale.from_factors([scale_factor] * 3)
         self.mesh.transform(S)
 
+    def scaled(self, scale_factor):
+        """Copies the collision mesh, and scales the copy uniformly.
+
+        Parameters
+        ----------
+        scale_factor : :obj:`float`
+            Scale factor.
+        """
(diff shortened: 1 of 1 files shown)


gonzalocasas:
LGTM!

gonzalocasas (approved):

PR #200, opened 2020-09-11
Use new extension point for declaring rhino installable packages

gonzalocasas:
This PR uses the changes on https://github.com/compas-dev/compas/pull/614 to remove the install/uninstall scripts and instead use the new extension points.

### Checklist

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -40,12 +40,11 @@ Working in Rhino
 ================
 
 To make **COMPAS FAB** available inside Rhino, open the *command prompt*
-and type the following which will install it on both Rhino 5.0 and 6.0:
+and type the following:
 
 ::
 
-    python -m compas_fab.rhino.install -v 5.0
-    python -m compas_fab.rhino.install -v 6.0
+    python -m compas_rhino.install
 
 .. note:
 
--- a/src/compas_fab/rhino/install.py
+++ b/src/compas_fab/rhino/install.py
@@ -2,26 +2,15 @@ from __future__ import absolute_import
 from __future__ import division
 from __future__ import print_function
 
-import compas_rhino.install
-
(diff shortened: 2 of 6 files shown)


src/compas_fab/rhino/install.py
```diff
@@ -2,26 +2,15 @@
 from __future__ import division
 from __future__ import print_function
 
-import compas_rhino.install
+import compas.plugins
```
beverlylytle, line 5: What version of compas has compas.plugins?  I think you might need to update the dependency.

gonzalocasas: good catch! I was actually a bit surprised this passes the build, but it's only because we don't test the rhino installer on CI.

The dependency needs to be updated, but right now, there's no release of `compas` containing this extension point. I was hesitating whether to add (a significant chuck of) code to handle releases before and after this extension point exists but it would only be useful for a very brief amount of time, so, I think it's best to just hold this PR until there's a release of compas, update the dependency there, and keep less code around.

gonzalocasas:
This should fix the issue mentioned above ☝️ 

@beverlylytle what do you think? ready to merge?

docs/getting_started.rst
```diff
@@ -40,12 +40,11 @@ Working in Rhino
...
 
 To make **COMPAS FAB** available inside Rhino, open the *command prompt*
-and type the following which will install it on both Rhino 5.0 and 6.0:
+and type the following:
 
 ::
 
-    python -m compas_fab.rhino.install -v 5.0
-    python -m compas_fab.rhino.install -v 6.0
+    python -m compas_rhino.install
```
beverlylytle, line 47: This will make compas_fab available?

beverlylytle:
Yep.

PR #201, opened 2020-09-11
close #165: rename RobotArtist to RobotModelArtist

### LINKED ISSUE #165: RobotArtist consuming RobotModel is not a great naming scheme
The current RobotArtist class is designed to consume a RobotModel object instead of a Robot object.

From naming consistency, user expect RobotArtist(Robot) and RobotModelArtist(RobotModel). Jsut like MeshArtist(Mesh), LineArtist(Mesh) ...

I don't have a good suggestion. But I am just getting confused about this all the time. Maybe there really should be two artist types.

gonzalocasas:
First step before moving these artists up to `compas` core, they've been rename to `RobotModel` artists.
This is a breaking change, although a minor one. I created aliases to preserve the old names for a little longer, but the `self.robot` instance variable of artists has changed to `self.model` and there's no alias for it.

### Checklist

### DIFF
--- a/src/compas_fab/blender/artists.py
+++ b/src/compas_fab/blender/artists.py
@@ -58,6 +58,5 @@ class RobotModelArtist(BaseRobotModelArtist):
         bpy.ops.wm.redraw_timer(type='DRAW_WIN_SWAP', iterations=1)
 
 
-
 # deprecated alias
 RobotArtist = RobotModelArtist
--- a/src/compas_fab/blender/artists.py
+++ b/src/compas_fab/blender/artists.py
@@ -4,7 +4,7 @@ from __future__ import print_function
 
 from compas_blender import draw_mesh
 
-from compas_fab.artists import BaseRobotArtist
+from compas_fab.artists import BaseRobotModelArtist
 
 try:
     import bpy
@@ -13,16 +13,24 @@ except ImportError:
     pass
 
 __all__ = [
-    'RobotArtist',
+    'RobotModelArtist',
 ]
 
 
-class RobotArtist(BaseRobotArtist):
-    """Visualizer for robots inside a Blender environment."""
+class RobotModelArtist(BaseRobotModelArtist):
+    """Visualizer for robot models inside a Blender environment.
 
-    def __init__(self, robot, layer=None):
+    Parameters
+    ----------
+    model : :class:`compas.robots.RobotModel`
+        Robot model.
+    layer : str, optional
+        The name of the layer that will contain the robot meshes.
+    """
+
+    def __init__(self, model, layer=None):
         self.layer = layer
-        super(RobotArtist, self).__init__(robot)
+        super(RobotModelArtist, self).__init__(model)
 
     def transform(self, native_mesh, transformation):
         native_mesh.matrix_world @= mathutils.Matrix(transformation.matrix)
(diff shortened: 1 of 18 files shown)


src/compas_fab/blender/artists.py
```diff
@@ -13,16 +13,24 @@
     pass
 
 __all__ = [
-    'RobotArtist',
```
beverlylytle, line 16: Shouldn't this stay?

gonzalocasas: It doesn't have to, but it could. The alias is explicitly imported in `blender/__init__.py` and in there, it is exposed in the `__all__`. But I can add it back in if you think it's less confusing.

beverlylytle: As you wish.

gonzalocasas:
oh.. sorry, some of these changes should not be here, they are update to run against the lastest `master` of compas (that's the error on integration tests now). I will revert them.

beverlylytle:
It seems like there should be an upper bound on the compas version, too.

gonzalocasas:
@beverlylytle sorry the delay in addressing the build issue here. It's now fixed to use the correct API if the version defined in setup's requirements (it is upper bounded already).

src/compas_fab/ghpython/path_planning.py
```diff
@@ -8,17 +8,15 @@
...
 from compas.datastructures import Mesh
 from compas.geometry import Transformation
-from compas_ghpython.geometry import xform_from_transformation
 from compas_rhino.geometry import RhinoMesh
+from compas_ghpython.geometry import xform_from_transformation
 
 from compas_fab.robots import Configuration
 
-try:
+if compas.RHINO:
```
beverlylytle, line 16: This doesn't seem to exist in compas 0.15.6.

src/compas_fab/rhino/artists.py
```diff
@@ -6,39 +6,38 @@
...
-
-from compas_fab.artists import BaseRobotArtist
-from compas.utilities import pairwise
 from compas.geometry import centroid_polygon
+from compas.utilities import pairwise
+
+from compas_fab.artists import BaseRobotModelArtist
 
-try:
+if compas.RHINO:
```
beverlylytle, line 14: Here, too.

src/compas_fab/ghpython/path_planning.py
gonzalocasas: Indeed, it doesn't, but this branch is already set to depend on `>=0.16.1,<0.17`: https://github.com/compas-dev/compas_fab/blob/rename_artists/setup.py#L14

beverlylytle: oh! strange, usually my ide complains when i don't meet requirements...

beverlylytle: But it also doesn't exist in 0.16.1 ;)

gonzalocasas: Damn! Semver broken! 🙀  Good catch, I bumped the min patch to be `0.16.2` at https://github.com/compas-dev/compas_fab/pull/201/commits/af671a4bd13a2218367d4879d7f4614b9b8cc4d0

beverlylytle (approved):

PR #203, opened 2020-09-18
Synchronous ROS planning scene methods

### LINKED ISSUE #95: Make planning scene methods synchronous
Currently, all planning scene modifications of the ROS backend (add/remove/append collision objects) are sent via topics, so there's no feedback about whether the operation was completed or not, leading to the proliferation of `time.sleep()` all around and the assumption that after waiting a bit, the operation will be finished.

However, ROS has a service to apply planning scene changes in deltas (diffs), which is probably much better for this case. We should switch the current implementation with a service-based one (which can also be used as blocking calls), to simplify user's code.

beverlylytle:
Closes #95 

### Checklist


beverlylytle:
Removing the advertising of topics and the initialization of the planner seems to really expose the problem with Twisted as the integration tests consistently fail for commit 066071b.  However, they pass on my machine.  Not sure what to do about this...

gonzalocasas (commented):
I know it's still draft, but this is looking good!

gonzalocasas:
In the stack trace, twisted errors out on a file called `_legacy.py` or similar. Could it be that twisted itself has changed and we cannot reproduce because we have slightly older versions of it locally?

beverlylytle:
I use Twisted version 20.3.0 locally, and that's what seems to be appearing in the logs for the integration tests.

beverlylytle:
I just created a new conda environment as is done in the install script (major difference being python 3.8 rather than 3.6) and ran the tests again, and they still pass locally.

gonzalocasas:
I checked what this legacy file is but it's only the logging system,so, unrelated.

One option to try to replicate would be to run the tests in an ubuntu docker container.

beverlylytle:
I've done this in a Debian container and I don't get the same errors.

conftest.py
```diff
@@ -0,0 +1,3 @@
+def pytest_configure(config):
+    from twisted.internet import selectreactor
+    selectreactor.install()
```
beverlylytle, line 3: The integration tests passed (this time) with this addendum, but locally when I ran the integration tests in a Debian container, they would mostly, but not always fail somewhere in the middle of `robot.py`.  I don't know if this chunk of code exposes, shifts the blame to, or introduces this.

gonzalocasas (approved):
[LGTM](https://lgtm.lol/i/431)

PR #212, opened 2020-09-23
Iter inverse kinematics

beverlylytle:
Here's a draft of the iterable inverse kinematics.  

### Checklist


beverlylytle:
I hereby open the discussion on what is desired of such a function.
In particular, there are a few TODOs scattered throughout that could be discussion points.
@gonzalocasas @romanarust @matteo-pacher

src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py
```diff
@@ -0,0 +1,91 @@
...
+            # random start configurations.
+            assert(link_name in robot.get_link_names(group=group))
+            sphere = Sphere(frame_WCF.point, max_reach)
+            constraints.append(PositionConstraint.from_sphere(link_name, sphere))
+        else:
+            # forcing diverse IK solutions by randomizing the start configuration
+            start_configuration = robot.random_configuration(group=group)
+
+        while True:
+            try:
```
matteo-pacher, line 84: I thought we wanted to avoid using any `try/except` construct. Is it needed?

src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py
```diff
@@ -0,0 +1,91 @@
...
+                yield self.ros_client.inverse_kinematics(robot,
+                                                         frame_WCF,
+                                                         start_configuration=start_configuration,
+                                                         group=group,
+                                                         options=options)
+            except BackendError:
+                yield "hello"  # TODO what should happen here?
```
matteo-pacher, line 91: I have no clue about this. Can it be the same as in the "normal" IK function?
```
compas_fab.backends.exceptions.BackendError
If no configuration can be found.
```

src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py
beverlylytle: There is probably a more elegant way to do this.  It was just an easy thing to do while I was making initial tests.

src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py
beverlylytle: If you don't catch the error, then it would make writing code like this next to useless:
```
for configuration in client.iter_inverse_kinematics(robot, frame, ...):
    # do something
```

src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py
```diff
@@ -0,0 +1,91 @@
...
+            - ``"link_name:``: (:obj:`str`, optional): TODO
+
+        Raises
+        ------
+        compas_fab.backends.exceptions.BackendError
+            If no configuration can be found.
+
+        Yields
+        -------
+        :obj:`tuple` of :obj:`list`
```
matteo-pacher, line 57: doesn't it yield a configuration?

src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py
```diff
@@ -0,0 +1,91 @@
...
+            # to generate his own constraints, and there could be a flag in the options on whether to generate
+            # random start configurations.
+            assert(link_name in robot.get_link_names(group=group))
+            sphere = Sphere(frame_WCF.point, max_reach)
+            constraints.append(PositionConstraint.from_sphere(link_name, sphere))
+        else:
+            # forcing diverse IK solutions by randomizing the start configuration
+            start_configuration = robot.random_configuration(group=group)
+
+        while True:
```
matteo-pacher, line 83: on my unit test, this `while` triggers an infinite loop, and never allows the yielding of any solution

beverlylytle: Right, you have to control how many iterations you want.  For example:
```
count = 0
num_iterations = 5
ik_gen = client.iter_inverse_kinematics(robot, frame, ...)
for config in ik_gen:
    # do something with config
    count += 1
    if count > num_iterations:
        break

File: src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py (hunk 1 above)
Comment (matteo-pacher) [reply src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py:85]: great! after you explained me, I appreciate the elegancy :)

File: src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py (hunk 2 above)
Comment (matteo-pacher) [reply src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py]: you're right

File: src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py (hunk 4 above)
Comment (matteo-pacher) [reply src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py:68]: my error. sorry

File: src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py [hunk 5]
```diff
@@ -0,0 +1,92 @@
...
+        constraints = options.setdefault('constraints', [])
+
+        group = group or robot.main_group_name
+
+        while True:
+
+            if max_reach and link_name:
+                # forcing diverse IK solutions by constraining the link_name within a sphere
+                #
+                # here I just want to bring up a discussion point:  since inverse_kinematics is called, and one might
```
matteo-pacher, line 73: that's a very good point and suggestion! I like it

romanarust: agree on that

src/compas_fab/backends/ros/backend_features/move_it_iter_inverse_kinematics.py
```diff
@@ -0,0 +1,92 @@
...
+from compas_fab.backends.interfaces import IterInverseKinematics
+from compas_fab.robots.constraints import PositionConstraint
+
+
+class MoveItIterInverseKinematics(IterInverseKinematics):
+    """Callable to iterate calculation of the robot's inverse kinematics for a given frame."""
+    def __init__(self, ros_client):
+        self.ros_client = ros_client
+
+    def iter_inverse_kinematics(self, robot, frame_WCF, start_configuration=None, group=None, options=None):
```
romanarust, line 17: one question remains open a bit if a user should expect `iter_inverse_kinematics` to return unique solutions.
in the case of analytic ik and 6-ax robot we can be sure to get at most 8, in the case of using moveit kdl and a robot on a linear axis we get infinite, and with moveit kdl only on a 6-ax-robot we could potentially get several close ones...

beverlylytle: I don't think that a user should expect that... As you suggest, each use case has a different set of outcomes, which can be influenced by the constraints a user provides.  I'm not sure how within the function call it would be known which of those various scenarios we are operating to thereby stop yielding configurations at the appropriate time.  And if you cached configurations within the generator to only yield unique configurations, I think it would be very easy to create an infinite (or by all appearances infinite) loop trying to find the next one.  I think we should expect the user to use heuristics about their particular robotic set-up, and use the iterator appropriately.

gonzalocasas:
@beverlylytle @matteo-pacher this PR is outdated, right? this feature is already in the latest release

PR #215, opened 2020-09-25
Compas update

beverlylytle:
Addresses changes in rhino related imports in compas v0.16.3/4.

### Checklist

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,7 @@ from setuptools import find_packages, setup
 
 requirements = [
     # Until COMPAS reaches 1.0, we pin major.minor and allow patch version updates
-    'compas>=0.16.2,<0.17',
+    'compas>=0.16.3,<0.17',
     'roslibpy>=1.1.0',
     'pybullet',
     'pyserial',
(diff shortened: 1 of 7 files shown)


beverlylytle:
Is there a better way to do this?

setup.py
```diff
@@ -11,7 +11,7 @@
 
 requirements = [
     # Until COMPAS reaches 1.0, we pin major.minor and allow patch version updates
-    'compas>=0.16.2,<0.17',
+    'compas>=0.16.3,<0.17',
```
beverlylytle, line 14: I think this is necessary because of the change of location of `xtransform` and `xform_from_transformation`.

gonzalocasas: Shouldn't we bump to `0.16.4`?

beverlylytle: Can do.

src/compas_fab/ghpython/path_planning.py
```diff
@@ -6,10 +6,7 @@
 from timeit import default_timer as timer
 
 import compas
-from compas.datastructures import Mesh
```
gonzalocasas, line 9: I'm so looking forward for the time when we can get rid of this file. I believe it's only used for vrep planning examples.

gonzalocasas:
I think it looks good, and I agree with the fix being temporary but necessary.

gonzalocasas (approved):
🚀 LGTM!

PR #221, opened 2020-10-16
Group state semantics

### LINKED ISSUE #156: Retrieve group states from `RobotSemantics`
# Feature Request

If srdf is available, it would be good to read group states (defined robot configurations).

beverlylytle:
Closes #156 

### Checklist

### DIFF
--- a/src/compas_fab/ghpython/artists.py
+++ b/src/compas_fab/ghpython/artists.py
@@ -9,7 +9,7 @@ if compas.RHINO:
     from compas_ghpython.utilities import draw_mesh
     from compas_rhino.geometry.transformations import xtransform
 
-from compas_fab.artists import BaseRobotModelArtist
+from compas_fab.artists import BaseRobotModelArtist  # noqa: E402
 
 __all__ = [
     'RobotArtist',
--- a/src/compas_fab/ghpython/artists.py
+++ b/src/compas_fab/ghpython/artists.py
@@ -4,12 +4,12 @@ from __future__ import print_function
 
 import compas
 from compas.utilities import rgb_to_rgb
+from compas_fab.artists import BaseRobotModelArtist
 
 if compas.RHINO:
     from compas_ghpython.utilities import draw_mesh
     from compas_rhino.geometry.transformations import xtransform
 
-from compas_fab.artists import BaseRobotModelArtist  # noqa: E402
 
 __all__ = [
     'RobotArtist',
--- a/src/compas_fab/robots/semantics.py
+++ b/src/compas_fab/robots/semantics.py
@@ -24,7 +24,8 @@ class RobotSemantics(object):
 
     def __init__(self, robot_model,
                  groups=None, main_group_name=None, passive_joints=None,
-                 end_effectors=None, disabled_collisions=None):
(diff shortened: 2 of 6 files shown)


src/compas_fab/robots/semantics.py
```diff
@@ -115,6 +119,18 @@ def _get_groups(root, robot_model):
...
 
 
+def _get_group_states(root):
+    group_states = {}
+
+    for group_state in root.findall('group_state'):
+        group_state_name = group_state.attrib['name']
+        group_name = group_state.attrib['group']
+        joint_dict = {joint.attrib['name']: joint.attrib['value'] for joint in group_state.findall('joint')}
+        group_states.setdefault(group_name, {})[group_state_name] = joint_dict
```
beverlylytle, line 129: Group state names are not unique (see panda.srdf), hence the nested dictionaries.  I also thought it would be more user friendly to have the group names as the top level keys rather than the group state names, but maybe others feel differently.

beverlylytle: @matteo-pacher

matteo-pacher: awesome! should I test in on the `rfl_description`?

beverlylytle: sure!

src/compas_fab/ghpython/artists.py
```diff
@@ -9,7 +9,7 @@
     from compas_ghpython.utilities import draw_mesh
     from compas_rhino.geometry.transformations import xtransform
 
-from compas_fab.artists import BaseRobotModelArtist
+from compas_fab.artists import BaseRobotModelArtist  # noqa: E402
```
gonzalocasas, line 12: why does the linter complain? is it because there's code above? would it be possible to move this above instead of noqa'ing the warning?

gonzalocasas (approved):
lgtm!

before merge, I'd wait for @matteo-pacher 's testing on the `rfl_description` model

src/compas_fab/robots/semantics.py
matteo-pacher: I tested the PR on `rfl_description` and all seems to work fine: group states can be retrieved properly, their joint names and values match the srdf information.
I wonder if it makes sense to expose the functionality in the `robot.py`, like it happens for `group_names` function.
I guess something like:

```
    @property
    def group_states(self):
        self.ensure_semantics()
        return self.semantics.group_states
```
but also maybe have a simple function to return a full configuration for a group_state?

src/compas_fab/ghpython/artists.py
beverlylytle: Yes, it's because of the code above.  I'll try to move it.

src/compas_fab/robots/semantics.py
beverlylytle: Thanks for testing this out.  Those are good ideas.  I will add the group_states property in this PR, but getting a full configuration from a group_state will be in another issue.

matteo-pacher: you're welcome!
you're right, I see how getting a full configuration from a group_state is maybe not a good idea.
Thank you Beverly!

PR #240, opened 2020-11-04
Two fixes in one

gonzalocasas:
This PR fixes two issues:
 - Attaching a tool was using a `frame_id=None` depending on the order of operations, so now we ensure the frame id is rebuilt if tool name changes
 - In some cases, IronPython uses an old-style class to represent the service response, which triggers an inner exception that is not handled and causes a freeze when used in GH.

### Checklist

### DIFF
--- a/src/compas_fab/robots/tool.py
+++ b/src/compas_fab/robots/tool.py
@@ -37,17 +37,16 @@ class Tool(object):
     def __init__(self, visual, frame_in_tool0_frame, collision=None,
                  name="attached_tool", link_name=None):
         self.tool_model = ToolModel(visual, frame_in_tool0_frame, collision, name, link_name)
-        self.attached_collision_meshes = self.get_attached_collision_meshes()
 
     @classmethod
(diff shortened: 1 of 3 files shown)


src/compas_fab/robots/tool.py
```diff
@@ -64,6 +65,7 @@ def link_name(self):
     @link_name.setter
     def link_name(self, link_name):
         self.tool_model.link_name = link_name
+        self.attached_collision_meshes = self.get_attached_collision_meshes()  # rebuild ACMs
```
beverlylytle, line 68: maybe this should just be a property that is built whenever it is accessed...

beverlylytle (approved):
LGTM!  The one comment can be addressed as you see fit.

src/compas_fab/robots/tool.py
gonzalocasas: moved it to a property on https://github.com/compas-dev/compas_fab/pull/240/commits/b47a15741ac172fb4cfad9a483de133917d2625e

PR #245, opened 2020-11-12
Pybullet add_attached_collision_mesh

beverlylytle:
Here's an attempt at fixing the pybullet add_attached_collision_mesh issues.  I ran into so much ugliness in the package `pybullet_utils` that I don't feel that this is truly mergeable.  But if we create our own `UrdfWriter`, there are some ideas in this code that could be used.  I create this as a draft PR to enable discussion, but I expect to close it while keeping the branch for later reference.


gonzalocasas:
Replaced by #259

PR #249, opened 2020-11-25
Feature/COMPAS FAB GH components

romanarust:

Draft for COMPAS FAB GH components made with the Python GHPY compiler.
Based on the [Tutorial by Giulio Piacentino](https://discourse.mcneel.com/t/tutorial-creating-a-grasshopper-component-with-the-python-ghpy-compiler/38552), which provides a transparent way of dealing with future versions of this components. 

For review, you have to run src\compas_fab\ghpython\components\compile.py from Rhino's PythonScriptEditor.
The library is then copied into 'USER\AppData\Roaming\Grasshopper\Libraries'

If you open Grasshopper, you'll see something like the following (icons need to be updated of course):
Capture

The robot_playground.ghx is updated with these components.

- [X] New feature in a **backwards-compatible** manner.

### Checklist


src/compas_fab/ghpython/components/assembly_info.py
```diff
@@ -0,0 +1,22 @@
+import GhPython
+import System
+
+assembly_name = "COMPAS FAB"
+assembly_version = "0.0.1"
```
gonzalocasas, line 5: maybe best here would be to use the same string we use for the overall `compas_fab` version. The way to do that is to 1) use the current version here, and 2) add this file and search/replace condition to the `.bumpversion.cfg` file on the root of the repo.

src/compas_fab/ghpython/components/compile.py
```diff
@@ -0,0 +1,31 @@
+import os
+from shutil import copyfile
+
+import clr
+
+assembly_name = "COMPAS FAB"
+assembly_version = "0.0.1"
```
gonzalocasas, line 7: same here with the version string

src/compas_fab/ghpython/components/compile.py
```diff
@@ -0,0 +1,31 @@
...
+                   "visualize_robot.py",
+                   "visualize_trajectory.py",
+                   )
+
+
+appdata = os.getenv('APPDATA')
+os.path.join(appdata, 'Grasshopper', 'Libraries')  # TODO: must move into compas_ghpython
+
+src = os.path.join(os.path.dirname(__file__), filename)
+dst = os.path.join(r'C:\Users\rustr\AppData\Roaming\Grasshopper\Libraries', filename)
```
gonzalocasas, line 30: I think we have functions in `compas_rhino.__init__` that can help you retrieve this folder dynamically.

src/compas_fab/ghpython/components/icons.py
```diff
@@ -0,0 +1,7 @@
+import System
+
+o = "iVBORw0KGgoAAAANSUhEUgAAABgAAAAYCAYAAADgdz34AAAAAXNSR0IArs4c6QAAAARnQU1BAACxjwv8YQUAAAAJcEhZcwAADsMAAA7DAcdvqGQAAAIlSURBVEhL7VM7aFRREN0mYqWIYLHkztzPW5UVQVw/+CNoYZNGC1EsRfxWFoIKQgqxUBDRThS0kbSJJqKCjSgIKdIFf6AWUQSttLCKZ+47q9n1hY1dihwYHnPmP3debRELH2vr9ZXJ+bGkfqa36Jvo4iaG9gYCds8Kvh5Uj0BOJpHTbSlUTxkP+wX4TZtvVD3HFHMjpeTMOTj/jNS8wEIzhYQDpKqBLkaCk19U/wtB/CMU+Vl4v5fUX6xxri4iKyw5ujmfnFuHYrei+EuQy+CuJAlHY4zLGfIPkHjApkji75MqgXW8LA2yEd/vQWQQ+72RuQqxogztAIqvos8EKSRX/y4Hqe43nQkuZmMXms3mEtiulf7+Nuk/sEtirtHsjI7vGmHroY8VfAFummolsK59THSCVAbO+o7xyfuDNZzZmezU1W0IYQuDr5KqBNY4jEIPqJo+yHxnMwHj81ytPxWZmIXo/XGz4R3e2upCf1iPaVv4bm54v9V0NLIaK9kZRfbA9yH97zFF3vWEkfYwpDqAhNth/2Y+VWLXVl5cqXevy+79Jg2HSFXC3gdd7sDE26x767ohEhuqG2y6qg1kFCJNPMoUdveJ1LxRFMUyxH3FvR8jVQ37mfKIIq9I9YR1XE7uf6jqUtJzI7iwq73H5GQSXzyYPoE8bQtW9BjTjsP+gQ29timYojdarVYfEg3hH3iPRJ9zok75CPkCmcSlHGbYIhY0arXfMz6/PVGPpY0AAAAASUVORK5CYII="
```
gonzalocasas, line 3: Since we have a build step, it would probably make sense to have these icon files as files and create the base64 encoded string at compile-time.

src/compas_fab/ghpython/components/ros_connect.py
```diff
@@ -0,0 +1,96 @@
...
+        if ros_client and (connect or disconnect):
+            ros_client.close()
+
+        if connect:
+            st[key] = RosClient(ip, 9090)
+            st[key].run(5)
+
+        ros_client = st.get(key, None)
+        is_connected = ros_client.is_connected if ros_client else False
+        return (ros_client, is_connected)
```
gonzalocasas, line 96: Isn't it more complicated to return this tuple instead of just `ros_client`, considering `is_connected` is also available on `ros_client.is_connected`?

src/compas_fab/ghpython/components/ros_robot.py
```diff
@@ -0,0 +1,82 @@
...
+            # Load URDF from ROS
+            loader = RosFileServerLoader(ros_client)
+            urdf = loader.load_urdf()
+            srdf = loader.load_srdf()
+
+            # Create robot model from URDF and load geometry
+            model = RobotModel.from_urdf_string(urdf)
+            model.load_geometry(loader)
+            semantics = RobotSemantics.from_srdf_string(srdf, model)
+            robot = Robot(model, semantics=semantics)
```
gonzalocasas, line 75: why all the explicit loading instead of doing `ros_client.load_robot(load_geometry=True)`?

gonzalocasas (commented):
Looks very good!! 
I added a couple of comments here and there, besides that, it seems the linter is complaining about some stuff, and I would add a short bit of documentation somewhere indicating the compilation process. Ideally also add a def in `tasks.py` so that it can be triggered like all other dev tasks.

src/compas_fab/ghpython/components/visualize_trajectory.py
```diff
@@ -0,0 +1,130 @@
...
+        p = Grasshopper.Kernel.Parameters.Param_GenericObject()
+        self.SetUpParam(p, "P", "P", "The positions along the trajectory")
+        self.Params.Output.Add(p)
+
+        p = Grasshopper.Kernel.Parameters.Param_GenericObject()
+        self.SetUpParam(p, "V", "V", "The velocities along the trajectory")
+        self.Params.Output.Add(p)
+
+        p = Grasshopper.Kernel.Parameters.Param_GenericObject()
+        self.SetUpParam(p, "A", "A", "The accellerations along the trajectory")
```
beverlylytle, line 74:
```suggestion
        self.SetUpParam(p, "A", "A", "The accelerations along the trajectory")
```

src/compas_fab/ghpython/components/ros_connect.py
romanarust: returning the bool as well is practical from a user perspective: you can simply add a textfield (panel) to that outlet, see below:

image

src/compas_fab/ghpython/components/assembly_info.py
romanarust: I changed bumpversion and added a few lines in 'CONTRIBUTING.rst'. But would need some help on adding a task and the naming of that task.
I created a file named install.py in src\compas_fab\ghpython\components\ that should be executed.

gonzalocasas:
@gonzalocasas ping myself. Add tasks or pluggable to install/copy components to Grasshopper folders when `compas_fab` is added to Rhino

gonzalocasas:
Idea for icons: use same compas icon + little label with feature (eg. `IK`, `FK`)

src/compas_fab/ghpython/components/ros_robot.py
```diff
@@ -6,6 +6,7 @@
 import compas
 from compas_fab.ghpython.components import create_id
 from compas_fab.ghpython.components.icons import ros_robot_icon
+from compas_fab.ghpython import RobotArtist
```
beverlylytle, line 9: This is still possible, but it's deprecated.  Should be 
```suggestion
from compas_ghpython.artists import RobotModelArtist
```

src/compas_fab/ghpython/components/ros_robot.py
```diff
@@ -60,6 +61,7 @@ def RunScript(self, ros_client, load):
         if ros_client and ros_client.is_connected and load:
             # Load URDF from ROS
             st[key] = ros_client.load_robot(load_geometry=True)
+            st[key].artist = RobotArtist(st[key].model)
```
beverlylytle, line 64:
```suggestion
            st[key].artist = RobotModelArtist(st[key].model)
```

src/compas_fab/ghpython/components/install.py
```diff
@@ -0,0 +1,38 @@
...
+from compas_fab.ghpython.components import filename
+
+# TODO: function below must move into compas_ghpython
+
+
+def _get_grasshopper_library_path():
+    if compas._os.system == 'win32':
+        grasshopper_library_path = os.path.join(os.getenv('APPDATA'), 'Grasshopper', 'Libraries')
+    elif compas._os.system == 'darwin':
+        raise NotImplementedError
```
beverlylytle, line 16: maybe this?
```suggestion
        grasshopper_library_path = os.path.join(
        os.getenv('HOME'), 'Library', 'Application Support', 'McNeel', 'Rhinoceros', '{}'.format(version), 
        'Plug-ins', 'Grasshopper (b45a29b1-4343-4035-989e-044e8580d9cf)', 'Libraries')
```

src/compas_fab/ghpython/components/install.py
```diff
@@ -0,0 +1,38 @@
...
+from shutil import copyfile
+import compas
+
+from compas_fab.ghpython.components import assembly_name
+from compas_fab.ghpython.components import filename
+
+# TODO: function below must move into compas_ghpython
+
+
+def _get_grasshopper_library_path():
```
beverlylytle, line 12: It might makes sense to add this function (one day)  to the `__init__.py` of `compas_ghpython`.

src/compas_fab/ghpython/components/assembly_info.py
```diff
@@ -0,0 +1,22 @@
...
+        return assembly_name
+
+    def get_AssemblyDescription(self):
+        return """Components for the COMPAS FAB package."""
+
+    def get_AssemblyVersion(self):
+        return assembly_version
+
+    def get_AuthorName(self):
+        return "rust@arch.ethz.ch"
```
gonzalocasas, line 19: Here you could use the generic `gramaziokohler@arch.ethz.ch` email, but also, you could do `from compas_fab.__version__ import __author_email__` and use that instead.

src/compas_fab/ghpython/components/attached_collision_mesh.py
```diff
@@ -0,0 +1,97 @@
...
+
+    def SetUpParam(self, p, name, nickname, description):
+        p.Name = name
+        p.NickName = nickname
+        p.Description = description
+        p.Optional = True
+
+    def RegisterInputParams(self, pManager):
+        p = Grasshopper.Kernel.Parameters.Param_GenericObject()
+        self.SetUpParam(p, "scene", "scene", ":class:`compas_fab.robots.PlanningScene` The planning scene.")
```
gonzalocasas, line 31: Do these strings really support the RST syntax of `:class:`?

src/compas_fab/ghpython/components/compile.py
```diff
@@ -0,0 +1,24 @@
+import clr
+
+# The below line does not work, since Grasshopper cannot be loaded from withing RhinoPython Editor.
+#from compas_fab.ghpython.components import filename
+
+assembly_name = "COMPAS FAB"
+assembly_version = "0.15.0"
```
gonzalocasas, line 7:
```suggestion
assembly_name = 'COMPAS FAB'
assembly_version = '0.15.0'
```

src/compas_fab/ghpython/components/compile.py
```diff
@@ -0,0 +1,24 @@
...
+filename = "%s_%s.ghpy" % (assembly_name, assembly_version)
```
gonzalocasas, line 8:
```suggestion
filename = '{}_{}.ghpy'.format(assembly_name, assembly_version)
```

src/compas_fab/ghpython/components/compile.py
```diff
@@ -0,0 +1,24 @@
+import clr
+
+# The below line does not work, since Grasshopper cannot be loaded from withing RhinoPython Editor.
+#from compas_fab.ghpython.components import filename
+
```
gonzalocasas, line 5: then maybe just remove it and also the comment. There's a bit of duplication but is not too bad, and the comment makes it feel like if something is broken and needs fixing.

src/compas_fab/ghpython/components/compile.py
```diff
@@ -0,0 +1,24 @@
...
+                   "ros_robot.py",
+                   "planning_scene.py",
+                   "collision_mesh.py",
+                   "attached_collision_mesh.py",
+                   "inverse_kinematics.py",
+                   "plan_cartesian_motion.py",
+                   "constraints_from_plane.py",
+                   "plan_motion.py",
+                   "visualize_robot.py",
+                   "visualize_trajectory.py",
```
gonzalocasas, line 23: Would it make sense to load these file names from disk instead of hard-coding them?

src/compas_fab/ghpython/components/plan_cartesian_motion.py
```diff
@@ -0,0 +1,114 @@
...
+        p7 = self.marshal.GetInput(DA, 7)
+        result = self.RunScript(p0, p1, p2, p3, p4, p5, p6, p7)
+
+        if result is not None:
+            self.marshal.SetOutput(result, DA, 0, True)
+
+    def get_Internal_Icon_24x24(self):
+        return plan_cartesian_motion_icon
+
+    def RunScript(self, robot, planes, start_configuration, group, attached_collision_meshes, path_constraints, max_step, compute):
```
gonzalocasas, line 94: While I know we do this in ghpython components, is the idea of having a `bool` to trigger long computation a common practice in other GH components?

src/compas_fab/ghpython/components/inverse_kinematics.py
```diff
@@ -0,0 +1,73 @@
...
+        if result is not None:
+            self.marshal.SetOutput(result, DA, 0, True)
+
+    def get_Internal_Icon_24x24(self):
+        return inverse_kinematics_icon
+
+    def RunScript(self, robot, plane, start_configuration, group):
+        configuration = None
+        if robot and robot.client and robot.client.is_connected and plane:
+            frame = Frame(plane.Origin, plane.XAxis, plane.YAxis)
```
gonzalocasas, line 71: This is a question that probably is to be left to a longer discussion outside this PR, but would it make sense to make all inputs support both compas and Rhino geometry so that they can be mixed and match? 

It would mean a lot of `isinstance` statements, but if we deem this useful, it could turn into a little utility function like `convert_to_compas_geometry()` that first checks if it already is a compas geometry object, and only if not, it does the conversion.

src/compas_fab/ghpython/components/ros_connect.py
```diff
@@ -0,0 +1,96 @@
...
+        port = port or 9090
+
+        key = create_id(self, 'ros_client')
+        ros_client = st.get(key, None)
+
+        if ros_client and (connect or disconnect):
+            ros_client.close()
+
+        if connect:
+            st[key] = RosClient(ip, 9090)
```
gonzalocasas, line 91:
```suggestion
            st[key] = RosClient(ip, port)
```

src/compas_fab/ghpython/components/ros_connect.py
```diff
@@ -0,0 +1,96 @@
...
+
+        p = Grasshopper.Kernel.Parameters.Param_Boolean()
+        self.SetUpParam(p, "connect", "connect", "If `True`, connect to ROS. Defaults to False.")
+        p.Access = Grasshopper.Kernel.GH_ParamAccess.item
+        self.Params.Input.Add(p)
+
+        p = Grasshopper.Kernel.Parameters.Param_Boolean()
+        self.SetUpParam(p, "disconnect", "disconnect", "If `True`, disconnect from ROS. Defaults to False.")
+        p.Access = Grasshopper.Kernel.GH_ParamAccess.item
+        self.Params.Input.Add(p)
```
gonzalocasas, line 50: Would it make sense to simplify this input into just one boolean that is intended to use with the toggle component? So, if `True` we connect, if `False`, disconnect?

src/compas_fab/ghpython/components/ros_robot.py
```diff
@@ -0,0 +1,69 @@
...
+        result = self.RunScript(p0, p1)
+
+        if result is not None:
+            self.marshal.SetOutput(result, DA, 0, True)
+
+    def get_Internal_Icon_24x24(self):
+        return ros_robot_icon
+
+    def RunScript(self, ros_client, load):
+        compas.PRECISION = '24f'
```
gonzalocasas, line 58: I normally use 12f without issues, any reasons for raising it to 24?

gonzalocasas (commented):
Looking good!!! I added a few comments here and there.
The `TODO` about the installation might need some addition to compas core. The current plugin to define rhino install packages only returns package names and compas does all the install. We don't yet have a way to hook on the install step and do something custom.
Alternatively, compas core could gain the ability to know how to install GH components instead, and then we would only need to provide the component locations. Both probably take a bit of dev time to get right. As a third alternative, we could add for now an entry-point like `compas_fab.ghpython.install` and do this completely in here. If we go that route, it would be a temp solution until we move it to core.

src/compas_fab/ghpython/components/install.py
romanarust: yes indeed ;)

src/compas_fab/ghpython/components/attached_collision_mesh.py
romanarust: no, it doesn't

src/compas_fab/ghpython/components/inverse_kinematics.py
romanarust: for the moment

src/compas_fab/ghpython/components/plan_cartesian_motion.py
romanarust: i don't know, but like that you are somewhat in control if you know that this component takes a bit long.

src/compas_fab/ghpython/components/ros_connect.py
romanarust: agree, updated

romanarust:
> Alternatively, compas core could gain the ability to know how to install GH components instead, and then we would only need to provide the component locations. 

I think that would be the way to go. I will open a draft PR on compas to get that started in parallel.

> As a third alternative, we could add for now an entry-point like `compas_fab.ghpython.install` and do this completely in here. If we go that route, it would be a temp solution until we move it to core.

Yes, this as temp solution that sounds right. Could you @gonzalocasas  add this?

src/compas_fab/ghpython/components/plan_cartesian_motion.py
gonzalocasas: 👍

gonzalocasas:
> Yes, this as temp solution that sounds right. Could you @gonzalocasas add this?

yep, will do

PR #256, opened 2020-12-13
Package update instructions. 

tetov:
Closes compas-dev/compas#697.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -36,6 +36,41 @@ Start Python from the command prompt and run the following:
 
 You are ready to use **COMPAS FAB**!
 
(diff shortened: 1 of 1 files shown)


gonzalocasas:
the deploy tasks fails because it's a PR from a fork and the GH_PAT is not exposed in that case, but that's alright

tetov:
Disable deploy if `$GITHUB_REPOSITORY != compas-dev/compas_fab`?

gonzalocasas:
> Disable deploy if $GITHUB_REPOSITORY != compas-dev/compas_fab?

perhaps, but whatever, not a big deal ;)

PR #265, opened 2021-02-10
add docker to contribution guide

beverlylytle:
As with roslibpy, I think it makes sense to add how to run the integration tests within the contribution guide.

### Checklist


beverlylytle:
oh, i suppose stopping the docker containers should also be listed...

gonzalocasas (approved):
The change in itself is 👍 
But since you at this, how about we add a single `docker-compose.yml` somewhere under `tests` that starts up MoveIt with the UR5 + VREP all in one go? I put together a first stab at that here:

[integration-tests.zip](https://github.com/compas-dev/compas_fab/files/5958427/integration-tests.zip)

beverlylytle:
Good idea.  Do you mean this docker compose file to be used for local testing or also for the github workflow?

gonzalocasas:
> Good idea. Do you mean this docker compose file to be used for local testing or also for the github workflow?

Ah good point! I thought about local testing, but this would make the github workflow faster (because it does not load the X forwarding part), so using it for both would be great

beverlylytle:
ok, @gonzalocasas, i added your docker-compose file, and it seems to work.  the integration tests are failing because of the `compas.get('faces.obj')` thing

gonzalocasas (approved):
🚀

PR #274, opened 2021-03-05
Dictionary-like Configuration

beverlylytle:
This is an attempt to make `Configuration` behave in some ways like a read-only dictionary, returning a joint value for a joint name key.  It is also an attempt to make the attributes like `joint_names` and `joint_values` semi-immutable in the sense that their lengths cannot change but their values can.  `FixedLengthList` subclasses `list` and overwrites many methods.  Please look over that carefully, and suggest meaningful tests.  That class is defined in `configuration.py` because it will only be there for a hot minute.  It too will migrate to compas core.  I haven't added an entry to the changelog for the same reason.  But at the same time, I wonder where best to announce the breaking change of `values` to `joint_values`, and the removal of `merge`.  

PS.  I just noticed that as part of the move I should probably update the PR template.

### Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -36,6 +36,7 @@ Unreleased
 **Changed**
 
 * Updated to ``COMPAS 1.1``
+* ``compas_fab.robots.JointTrajectoryPoint.values`` became ``compas_fab.robots.JointTrajectoryPoint.joint_values``
 
 **Fixed**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -31,7 +31,7 @@ Unreleased
 * Added serialization methods to ``compas_fab.robots.CollisionMesh`` and ``compas_fab.robots.AttachedCollisionMesh``
 * Added ``attached_collision_meshes`` attribute to ``compas_fab.robots.JointTrajectory``
 * Added ``compas_fab.backends.ros.PlanningSceneComponents.__ne__``
-* Added read-only dictionary behave to ``compas_fab.robots.JointTrajectoryPoint.merge``
+* Added dictionary behavior to ``compas_fab.robots.JointTrajectoryPoint.merge``
 * Added length limitations to attributes of ``compas_fab.robots.JointTrajectoryPoint.merge``
 
 **Changed**
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -37,7 +37,7 @@ Unreleased
 **Changed**
 
 * Updated to ``COMPAS 1.1``
-* ``compas_fab.robots.JointTrajectoryPoint.values`` became ``compas_fab.robots.JointTrajectoryPoint.joint_values``
+* ``Configuration`` & ``JointTrajectoryPoint``: the attributes ``values`` and ``types`` changed to ``joint_values`` and `joint_types` respectively.
 
 **Fixed**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -30,6 +30,8 @@ Unreleased
 * Added ``compas_fab.robots.Robot.ensure_geometry``
 * Added serialization methods to ``compas_fab.robots.CollisionMesh`` and ``compas_fab.robots.AttachedCollisionMesh``
 * Added ``attached_collision_meshes`` attribute to ``compas_fab.robots.JointTrajectory``
+* Added read-only dictionary behave to ``compas_fab.robots.JointTrajectoryPoint.merge``
+* Added length limitations to attributes of ``compas_fab.robots.JointTrajectoryPoint.merge``
 
 **Changed**
 
@@ -44,6 +46,8 @@ Unreleased
 **Deprecated**
 
 **Removed**
+
+* Remove ``compas_fab.robots.JointTrajectoryPoint.merge``
 
 0.16.0
 ----------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -7,6 +7,19 @@ All notable changes to this project will be documented in this file.
 The format is based on `Keep a Changelog <https://keepachangelog.com/en/1.0.0/>`_
 and this project adheres to `Semantic Versioning <https://semver.org/spec/v2.0.0.html>`_.
 
+Unreleased
+----------
+
+**Added**
+
+**Changed**
+
+**Fixed**
+
+**Deprecated**
+
+**Removed**
+
 0.17.0
 ----------
 
--- a/src/compas_fab/robots/configuration.py
+++ b/src/compas_fab/robots/configuration.py
@@ -3,7 +3,6 @@ from __future__ import division
 from __future__ import print_function
 
 from math import pi
-from collections import UserDict
 from compas.robots import Joint
 
 __all__ = [
@@ -13,6 +12,8 @@ __all__ = [
 
 class Configuration(object):
     """Represents the configuration of a robot based on the state of its joints.
+    If the names of joints are also provided, the configuration behaves as a
+    read-only dictionary of joint name-value pairs.
 
     This concept is also refered to as \"Joint State\".
 
@@ -95,6 +96,15 @@ class Configuration(object):
                 return value
         raise KeyError(item)
 
+    def __bool__(self):
+        return bool(self.joint_values)
+
+    def __len__(self):
+        return len(self.joint_names)
+
+    def __iter__(self):
+        return iter(self.joint_names)
+
     def items(self):
         return zip(self.joint_names, self.joint_values)
 
@@ -103,6 +113,13 @@ class Configuration(object):
 
     def values(self):
         return iter(self.joint_values)
+
+    def get(self, key, default=None):
+        try:
+            value = self[key]
+        except KeyError:
+            value = default
+        return value
 
     @classmethod
     def from_revolute_values(cls, values, joint_names=None):
--- a/src/compas_fab/robots/configuration.py
+++ b/src/compas_fab/robots/configuration.py
@@ -17,7 +17,7 @@ class Configuration(object):
 
     Parameters
     ----------
-    values : :obj:`list` of :obj:`float`
+    joint_values : :obj:`list` of :obj:`float`
         Joint values expressed in radians or meters, depending on the respective
         type.
     types : :obj:`list` of :attr:`compas.robots.Joint.SUPPORTED_TYPES`
@@ -28,7 +28,7 @@ class Configuration(object):
 
     Attributes
     ----------
-    values : :obj:`list` of :obj:`float`
+    joint_values : :obj:`list` of :obj:`float`
         Joint values expressed in radians or meters, depending on the respective
         type.
     types : :obj:`list` of :attr:`compas.robots.Joint.SUPPORTED_TYPES`
@@ -46,7 +46,7 @@ class Configuration(object):
     Examples
     --------
     >>> config = Configuration.from_revolute_values([math.pi/2, 0., 0.])
-    >>> config.values
+    >>> config.joint_values
     [1.5707963267948966, 0.0, 0.0]
 
     >>> from compas_fab.robots import Configuration
@@ -62,19 +62,19 @@ class Configuration(object):
 
     """
 
-    def __init__(self, values=None, types=None, joint_names=None):
+    def __init__(self, joint_values=None, types=None, joint_names=None):
         self._precision = '3f'
-        self.values = list(values or [])
+        self.joint_values = list(joint_values or [])
         self.types = list(types or [])
         self.joint_names = list(joint_names or [])
 
-        if len(self.values) != len(self.types):
-            raise ValueError("%d values must have %d types, but %d given." % (
-                len(self.values), len(self.values), len(self.types)))
+        if len(self.joint_values) != len(self.types):
+            raise ValueError("%d joint_values must have %d types, but %d given." % (
+                len(self.joint_values), len(self.joint_values), len(self.types)))
 
     def __str__(self):
         """Return a human-readable string representation of the instance."""
-        v_str = ('(' + ", ".join(['%.' + self._precision] * len(self.values)) + ')') % tuple(self.values)
+        v_str = ('(' + ", ".join(['%.' + self._precision] * len(self.joint_values)) + ')') % tuple(self.joint_values)
         if len(self.joint_names):
             return "Configuration({}, {}, {})".format(v_str, tuple(self.types), tuple(self.joint_names))
         else:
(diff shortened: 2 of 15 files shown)


tests/robots/test_configuration.py
```diff
@@ -6,27 +7,43 @@
 from compas_fab.robots import to_degrees
 
 
+def test_bool():
+    config = Configuration(joint_values=[1, 2, 3], types=[Joint.REVOLUTE]*3, joint_names=['a', 'b', 'c'])
+    assert config
+
+    config = Configuration(joint_values=[1, 2, 3], types=[Joint.REVOLUTE]*3)
+    assert config
```
gonzalocasas, line 15: A test for `False` would also make sense (assuming we agree that `Configuration(joint_values=[0, 0, 0])` is really intended to be assumed as `False` when evaluated as bool)

tests/robots/test_configuration.py
```diff
@@ -6,27 +7,43 @@
...
+
+
+def test_len():
+    config = Configuration(joint_values=[1, 2, 3], types=[Joint.REVOLUTE]*3, joint_names=['a', 'b', 'c'])
+    assert len(config) == 3
+
+    config = Configuration(joint_values=[1, 2, 3], types=[Joint.REVOLUTE]*3)
+    assert len(config) == 0
```
gonzalocasas, line 23: I don't like this much. I understand why it happens, but it does not make much sense, especially if you consider that names are optional. Should we automatically assign `joint_i` names if joints are not passed? Or would that break stuff?

src/compas_fab/robots/configuration.py
```diff
@@ -84,6 +148,41 @@ def __repr__(self):
         """Printable representation of :class:`Configuration`."""
         return self.__str__()
 
+    def __getitem__(self, item):
+        for name, value in zip(self.joint_names, self.joint_values):
+            if name == item:
+                return value
+        raise KeyError(item)
+
```
gonzalocasas, line 156: Please remind me why aren't we adding a `__setitem__`? I think we talked about it, but now it feels like it's totally missing and I cannot recall why it would be problematic (maybe related to optional joint names, but in that case, getter accessor doesn't work either). I tested this and superficially seems to work:

```suggestion

    def __setitem__(self, item, value):
        i = self.joint_names.index(item)
        if i < 0:
            raise KeyError(item)
        self.joint_values[i] = value

```

tests/robots/test_configuration.py
```diff
@@ -56,22 +73,44 @@ def test_to_data():
...
+    assert config['a'] == 4
+    assert config.get('a') == 4
+    assert config.get('DNE') is None
+    assert config.get('DNE', 5) == 5
+    with pytest.raises(TypeError):
+        config['a'] = 1
+    with pytest.raises(TypeError):
+        config.joint_names.append('d')
+    with pytest.raises(TypeError):
+        config.types[1:1] = range(5)
```
gonzalocasas, line 115: i'd suggest to split some of these checks to give them a name, for example, here something like `test_length_immutability` or similar).

tests/robots/test_configuration.py
```diff
@@ -56,22 +73,44 @@ def test_to_data():
...
+    assert list(config.items()) == [('a', 1), ('b', 2), ('c', 3)]
+    assert list(config.keys()) == ['a', 'b', 'c']
+    assert list(config.values()) == [1, 2, 3]
+    with pytest.raises(KeyError):
+        _ = config['DNE']
+    assert config['a'] == 1
+    config.joint_values[0] = 4
+    assert config['a'] == 4
+    assert config.get('a') == 4
+    assert config.get('DNE') is None
```
gonzalocasas, line 108: just to make sure we don't break this in the future, please add a test that does uses get with default on a joint that has `0` as value, something like:
```python
config = Configuration([0, 1, 2], [0, 0, 0], ['a', 'b', 'c'])
assert config.get('a', 5) == 0
```

src/compas_fab/robots/configuration.py
```diff
@@ -62,19 +92,53 @@ class Configuration(object):
 
     """
 
-    def __init__(self, values=None, types=None, joint_names=None):
+    def __init__(self, joint_values=None, types=None, joint_names=None):
+        joint_values = FixedLengthList(joint_values or [])
+        types = FixedLengthList(types or [])
+        joint_names = FixedLengthList(joint_names or [])
```
gonzalocasas, line 98: It's currently possible to pass a list of `joint_names` with repeated names.
Here, Python is inconsistent is that, depending on the syntax you use, it will allow this or not, ie. `dict(a=1, a=3)` is a `SyntaxError`, but `{'a':1, 'a':3}` is last-one-wins. I find an error much better because it prevents a whole lot of other mistakes, so, I would check and raise an exception here.

gonzalocasas (changes requested):
It's looking good but there are a number of cases that need fixing!  Subclassing list to fix length is bold, I keep wondering if there's no built-in option anywhere for this (there isn't, afaik, but it just feels odd) and/or if we should immediately move this upstream. However, since we are upstreaming this change asap anyway, I guess the latter part is answered. ;)

src/compas_fab/robots/configuration.py
```diff
@@ -62,19 +92,53 @@ class Configuration(object):
...
+
+        if len(joint_values) != len(types):
+            raise ValueError("%d joint_values must have %d types, but %d given." % (
```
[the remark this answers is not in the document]
beverlylytle: I agree.  I just copied this from somewhere else.  Oh, to be able to use f-strings....

src/compas_fab/robots/configuration.py
```diff
@@ -62,19 +92,53 @@ class Configuration(object):
...
+        self._types = types
+        self._joint_names = joint_names
+
+    @property
+    def joint_values(self):
+        return self._joint_values
+
+    @joint_values.setter
+    def joint_values(self, values):
+        if len(self._joint_values) != len(values):
```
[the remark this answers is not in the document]
beverlylytle: Do you find the docstring for `Configuration` to be insufficient?  I find it a bit strange that `@property` defined attributes are separated from the `__init__` defined attributes in the documentation.

src/compas_fab/robots/configuration.py
```diff
@@ -84,6 +148,41 @@ def __repr__(self):
...
         return self.__str__()
 
+    def __getitem__(self, item):
+        for name, value in zip(self.joint_names, self.joint_values):
+            if name == item:
+                return value
+        raise KeyError(item)
+
+    def __bool__(self):
+        return bool(self.joint_values)
```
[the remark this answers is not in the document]
beverlylytle: Arg.  There was a very specific reason I added this, and I cannot remember now.  

I'm not sure what you mean.  The results of `robot.zero_configuration()` would have `joint_values` set to `[0, ..., 0]`, which is not falsy.  But an empty configuration with no joint values or types, but (in this implementation) potentially names... how would that cause problems if that were falsy?

tests/robots/test_configuration.py
beverlylytle: Oh, no, `Configuration(joint_values=[0, 0, 0])` is not `False`.

beverlylytle: potentially `Configuration(joint_names=['a', 'b', 'c'])` is `False`.

src/compas_fab/robots/configuration.py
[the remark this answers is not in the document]
gonzalocasas: not insufficient but if one navigates directly to the props, then they are empty: 
image

src/compas_fab/robots/configuration.py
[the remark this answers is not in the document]
gonzalocasas: aha! sorry, I incorrectly assumed that a list of zeros was also falsy. then, dismiss this one! ;)

tests/robots/test_configuration.py
beverlylytle: I think that would break stuff.  There are definitely checks here and there on whether a configuration has names attached.  And if the user hasn't updated those names to the real things...

for example in `robot.merge_group_with_full_configuration`....

gonzalocasas: Ok, then let's not do that ;)
But I still find it unexpected/exotic that `bool` is defined in terms of values but `len` is defined in terms of names. As a user, you have no way of guessing this difference unless you read the code. Would it be an option to define `len` in terms of values instead and maybe raise an exception if `len(joint_values) != len(joint_names)`?

src/compas_fab/robots/configuration.py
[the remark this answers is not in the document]
beverlylytle: ok, can do

src/compas_fab/robots/configuration.py
beverlylytle: This is a very good point.

src/compas_fab/robots/configuration.py
beverlylytle: I think the reason had to do with user experience and expectations and being consistent with being a "read-only" dictionary rather than the worry of things breaking.  If this is added, then you can write things to the dictionary, but only sometimes.  Without knowing the internals, maybe it would be confusing to a user which aspects of a dictionary does a configuration have and which doesn't it.  "read-only" is succinct, but limiting.  "readable, and also over-writable, but not generally writable" is a bit much to explain.

tests/robots/test_configuration.py
beverlylytle: I feel pretty strongly that `len` should reflect the number of elements encountered while iterating over the object, in this case, the `joint_names`.  and i don't like the idea of `__len__` raising an exception.  but i am at a loss as to why i decided `bool` should be determined by `joint_values`.  there is no test (that i didn't write) that fails for this.  maybe `__bool__` should be left undefined.

src/compas_fab/robots/configuration.py
gonzalocasas: ...but currently is not read-only either, it's only fixed-length but values can be changed (otherwise we could use tuples instead of `FixedLengthList`)

beverlylytle: yes, true, but... the writability is not a function of its dictionary-like qualities.  in any case, i don't mind adding the `__setitem__`.

tests/robots/test_configuration.py
gonzalocasas: agree that it makes more sense than `__len__` is based on `joint_names` because it's part of the dict interface. Undefining `__bool__` would be an option if you don't rely on that behavior for anything else.

beverlylytle: oh, `if not config` is used in the vrep client `set_robot_config`, so leaving `__bool__` undefined fails there.  should it always return `True`?

gonzalocasas: but maybe those usages could be replaced for `if config is None`... at least a quick look at both `set_robot_config` and `set_robot_pose` seem to indicate both are intended to be null-checks.

beverlylytle: oh, right, of course, if `__bool__` is not defined, then the default `__bool__` calls `__len__`, so that's why i did it the way i did... because if `joint_names` is empty, that doesn't mean the `config` should be falsy.

beverlylytle: So I think it makes the most sense to have `__bool__` always return `True` which is the default behavior when there is no overwriting of `__bool__` and `__len__`.

beverlylytle: and then i don't have to worry about breaking things in the v-rep client or elsewhere.

gonzalocasas: 👍 I would definitely add tests and an inline comment about why these two return what they do (as summary of this thread basically), otherwise it's a given that we'll forget and break it next time we touch it.

src/compas_fab/robots/configuration.py
beverlylytle: is it ok to ignore the possibility that a user might do something like `config.joint_names[0] = config.joint_names[1] = 'a'`? otherwise modifying `FixedLengthList` to include optional validation in `__setitem__` and `__setslice__` would be really tricky.

gonzalocasas: yep 👍

beverlylytle:
I added a validator to `FixedLengthList`.  Maybe this was a horrible idea....

src/compas_fab/robots/configuration.py
```diff
@@ -62,28 +110,132 @@ class Configuration(object):
...
+        if len(joint_values) != len(joint_types):
+            raise ValueError('{} joint_values must have {} joint_types, but {} given.'.format(
+                len(joint_values), len(joint_values), len(joint_types)))
+
+        if joint_names and len(joint_names) != len(joint_values):
+            raise ValueError('{} joint_values must have either 0 or {} names, but {} given'.format(
+                len(joint_values), len(joint_values), len(joint_names)
+            ))
+
         self._precision = '3f'
```
beverlylytle, line 127: Should this actually be `compas.PRECISION`?

gonzalocasas: Is it used for anything other than the `__str_`? If that's the only usage, then, no need to.

CHANGELOG.rst
```diff
@@ -31,10 +31,13 @@ Unreleased
 * Added serialization methods to ``compas_fab.robots.CollisionMesh`` and ``compas_fab.robots.AttachedCollisionMesh``
 * Added ``attached_collision_meshes`` attribute to ``compas_fab.robots.JointTrajectory``
 * Added ``compas_fab.backends.ros.PlanningSceneComponents.__ne__``
+* Added dictionary behavior to ``compas_fab.robots.JointTrajectoryPoint.merge``
+* Added length limitations to attributes of ``compas_fab.robots.JointTrajectoryPoint.merge``
 
 **Changed**
 
 * Updated to ``COMPAS 1.1``
+* ``compas_fab.robots.JointTrajectoryPoint.values`` became ``compas_fab.robots.JointTrajectoryPoint.joint_values``
```
gonzalocasas, line 40:
```suggestion
* ``Configuration`` & ``JointTrajectoryPoint``: the attributes `values` and `types` changed to `joint_values` and `joint_types` respectively.

File: src/compas_fab/robots/robot.py [hunk 12]
```diff
@@ -211,8 +211,7 @@ def get_end_effector_frame(self, group=None, full_configuration=None):
         """
         if not full_configuration:
             full_configuration = self.zero_configuration()
-        full_joint_state = dict(zip(full_configuration.joint_names, full_configuration.values))
-        return self.model.forward_kinematics(full_joint_state, link_name=self.get_end_effector_link_name(group))
+        return self.model.forward_kinematics(full_configuration, link_name=self.get_end_effector_link_name(group))
```
gonzalocasas, line 214: 💪

gonzalocasas (approved):
LGTM!! 
LGTM

src/compas_fab/robots/robot.py
beverlylytle: Yeah, it's nice, isn't it?

beverlylytle:
aww, you're just being nice... that or you are tired of looking at this ;)

PR #288, opened 2021-04-01
Introducing PartialOrder

### LINKED ISSUE #202: Add partial order plan datastructure

beverlylytle:
This pull request is to introduce the new second-level module `compas_fab.datastructures` and its constituents `PartialOrder` and `Action`.  These come together to form a partially ordered plan. Closes #202 

### Checklist


beverlylytle:
I thought I'd open a draft pull request to get commentary on choices I've made so far, before I start adding docstrings and tests.  First big change is naming.  Second is decoupling the way the ids are generated (and their values) from anything in the `Plan`.  Third, I've added a method to remove planned actions.  I'm debating the necessity of `namespace` and thinking of removing it unless otherwise convinced.  I was thinking also of adding the methods `check_for_cycles` and `sorted`.  Or is that premature optimization?

yck011522:
@beverlylytle  do you mind short elaborate what the Action class and Plan class is supposed to hold? 

Is an Action a movement waiting to be path planned? Is it a supposed to hold like a initial config and target config?

src/compas_fab/datastructures/plan.py
```diff
@@ -0,0 +1,130 @@
...
+
+__all__ = [
+    'Action',
+    'IntegerIdGenerator',
+    'Plan',
+    'PlannedAction',
+]
+
+
+class IntegerIdGenerator(object):
```
gonzalocasas, line 20: Would it make sense to make this thread-safe, like the `SequenceCounter` in rrc?

beverlylytle:
> Is an Action a movement waiting to be path planned? Is it a supposed to hold like a initial config and target config?

This whole thing is supposed to be very generic.  An `Action` can hold whatever the user wants in `parameters`.  It could be the location and mesh of brick, and then the `Plan` could represent the assembly sequence.  Or it could be that the `Action` holds a `ROSmsg` and then the `Plan` represents the fabrication schedule.  It really is suppose to hold the data for anything that can be partially ordered.

src/compas_fab/datastructures/plan.py
beverlylytle: yep

src/compas_fab/datastructures/plan.py
```diff
@@ -0,0 +1,130 @@
...
+        self.namespace = namespace
+
+    def __str__(self):
+        return 'PlannedAction<id={}, action={}, namespace={}>'.format(self.id, self.action, self.namespace)
+
+    @property
+    def data(self):
+        return dict(
+            id=self.id,
+            action=self.action.to_data(),
```
gonzalocasas, line 94: this nested `.to_data()` comes from the pre-recursive decoders (we probably still have a few of these in other parts of the code). For new classes, we can keep the data with "live" objects, and the dataencoder/datadecoder will handle properly. Also the `from/to_data` methods should not be needed if inheriting from the base datastructures of core.

yck011522:
> > Is an Action a movement waiting to be path planned? Is it a supposed to hold like a initial config and target config?
> 
> This whole thing is supposed to be very generic. An `Action` can hold whatever the user wants in `parameters`. It could be the location and mesh of brick, and then the `Plan` could represent the assembly sequence. Or it could be that the `Action` holds a `ROSmsg` and then the `Plan` represents the fabrication schedule. It really is suppose to hold the data for anything that can be partially ordered.

Sorry, I don't fully get it, can you maybe give the intended use example you have in mind for the three classes Action, PlannedAction and Plan. I guess Plan holes an ordered list of PlannedAction? But the use case of Action is not so clear to me. Do I start with a Action. and pass it to a planner somehow, and I will get a PlannedAction in return? And how is the dependency_ids used? Is it like a tree structure for seeking out dependent Actions that are not planned and plan those first?

I dont see any information how these classes connect with the planners, so amybe I'm completely guessing wrong.

beverlylytle:
Ah, sorry, sure.  Yeah, this has nothing to do with any of the planners.  This is purely a data structure for holding a plan, and  with no algorithmic planning capacity.

So, imagine you have three `Brick`s (`bricks = [brick_1, brick_2, brick_3]`) and these bricks are to be stacked into a pyramid with a base of `brick_1` and `brick2` and with `brick_3` on top.  You could model the assembly sequence with `Plan` by creating an action for each brick:
```
actions = {}
for i, brick in enumerate(bricks):
    name = 'brick_{}'.format(i)
    parameters = {'brick': brick}
    action = Action(name, parameters)
    actions[name] = action
```
Now, we need to plan these actions.  The idea is that some `Action`s should "occur" before other `Action`s, but they don't necessarily form a linear sequence.  For example, the first two bricks can be place in either order or even simultaneously.  Since they are dependent on nothing, they get empty `dependency_ids`.
```
plan = Plan()
brick_1_action_id = plan.plan_action(actions['brick_1'], [])
brick_2_action_id = plan.plan_action(actions['brick_2'], [])
```
But the third brick sits on top of the other two, so for it to be a sensical assembly sequence, they would have to be placed before this third brick.  So the third brick depends on the the first two.
```
brick_3_action_id = plan.plan_action(actions['brick_3'], [brick_1_action_id, brick_2_action_id])
```
Now we have a `Plan` with three `PlannedAction`s that were created with `plan_action` our `Action`s.   This can be thought of as a graph with the `PlannedAction`s forming the nodes, an `Action` being the content of a node and the `dependency_ids` pointing to the parents of the node.   This is not a tree because there could be multiple "roots" and could be disconnected, but it should be acyclic.  (The current implementation doesn't enforce that, but that is the intent of this structure.)  So in the above example, brick 3 points to bricks 1 and 2.  This gives a partial order on the three bricks, which can be sorted into a linear order, but doesn't have to be.  

It would also be possible to traverse this dependency graph (using threading) to complete some tasks held in each action (say, emit some ros messages) in a partially ordered manner.  So in the above example, ros messages could be attached to each  brick's action for picking and placing that brick.  Traversing the graph, messages would be emitted for the picking and placing of bricks 1 and 2, allowing for parallel movement of two robots, and then the ros messages for picking and placing brick 3 wouldn't start until feedback had been received that the tasks for actions 1 and 2 were complete.

Does that make more sense @yck011522 ?

src/compas_fab/datastructures/plan.py
beverlylytle: Is this new version better?

src/compas_fab/datastructures/plan.py
```diff
@@ -0,0 +1,368 @@
...
+    added to the plan through a :class:`compas_fab.datastructures.PlannedAction`.
+    The dependency ids of a planned action can be thought of as pointers to the
+    parents of that planned action.
+
+    Parameters
+    ----------
+    planned_actions : list of :class:`compas_fab.datastructures.PlannedAction`
+        The planned actions will be stored as an ordered dictionary, so their
+        ids should be distinct.  Defaults to an empty list.
+    id_generator : Generator[Hashable, None, None]
```
beverlylytle, line 97: Unsure about the typing here, and how sphinx wants it...

src/compas_fab/datastructures/plan.py
```diff
@@ -0,0 +1,368 @@
...
+        ----------
+        action : :class:`comaps_fab.datastructures.Action`
+            The action to be added to the plan.
+        dependency_ids : set or list
+            The keys of the already planned actions that the new action
+            is dependent on.
+
+        Returns
+        -------
+            The id of the newly planned.
```
beverlylytle, line 137:
```suggestion
            The id of the newly planned action.
```

src/compas_fab/datastructures/plan.py
```diff
@@ -0,0 +1,368 @@
...
+        to the plan, and generates an id for this newly planned action.
+
+        Parameters
+        ----------
+        action : :class:`comaps_fab.datastructures.Action`
+            The action to be added to the plan.
+
+        Returns
+        -------
+            The id of the newly planned.
```
beverlylytle, line 156:
```suggestion
            The id of the newly planned action.
```

src/compas_fab/datastructures/plan.py
```diff
@@ -0,0 +1,368 @@
...
+            dependency_ids = {last_action_id}
+        return self.plan_action(action, dependency_ids)
+
+    def remove_action_by_id(self, action_id):
+        """Removes the action with the given id from the plan and all its
+        dependencies.
+
+        Parameters
+        ----------
+        action_id : Hashable
```
beverlylytle, line 170: Similarly here.  What type should this have?

src/compas_fab/datastructures/plan.py
```diff
@@ -0,0 +1,368 @@
...
+            Id of the planned action to be removed.
+
+        Returns
+        -------
+        :class:`compas_fab.datastructure.Action`
+        The action of the removed :class:`compas_fab.datastructure.PlannedAction`
```
beverlylytle, line 176:
```suggestion
            The action of the removed :class:`compas_fab.datastructure.PlannedAction`
```

gonzalocasas (commented):
I'm leaving some initial feedback, but I'm yet to test this on the project that cannot be named! ;)

src/compas_fab/datastructures/plan.py
```diff
@@ -0,0 +1,371 @@
...
+        )
+
+    @data.setter
+    def data(self, data):
+        self.planned_actions = data['planned_actions']
+        self._id_generator = data['id_generator']
+
+    @classmethod
+    def from_data(cls, data):
+        return cls(**data)
```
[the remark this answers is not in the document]
beverlylytle: That requires that I provide default values for all of the constructor arguments, which I didn't want to do because I wanted to force the user to provide them.

src/compas_fab/datastructures/plan.py
```diff
@@ -0,0 +1,371 @@
...
+        -------
+            The id of the newly planned action.
+        """
+        self.check_dependency_ids(dependency_ids)
+        action_id = self._get_next_action_id()
+        planned_action = PlannedAction(action_id, action, dependency_ids)
+        self.planned_actions[action_id] = planned_action
+        return action_id
+
+    def append_action(self, action):
```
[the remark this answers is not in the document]
beverlylytle: I wanted to make a bigger distinction between `dependency_ids=None` and `dependency_ids=[]` because they result in very different behavior and I can foresee that being the source of annoying bugs for the user.

src/compas_fab/datastructures/plan.py
```diff
@@ -0,0 +1,371 @@
...
+    def data(self, data):
+        self.planned_actions = data['planned_actions']
+        self._id_generator = data['id_generator']
+
+    @classmethod
+    def from_data(cls, data):
+        return cls(**data)
+
+
+class PlannedAction(Datastructure):
```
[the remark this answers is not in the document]
beverlylytle: I am unaware of the compas definition of data structure.  It seems to me though, that `PlannedAction` and `Action` are exactly objects to hold user data, and not some class to hold functionality like, I don't know, `LazyLoader`.  They are also definitely not `geometry` and not `robots`.  But I suppose they do violate the compas assumption that data structures have empty constructors. So I will move subclass from `Base` instead, and copy some of the overwritten methods from `Datastructure`.

src/compas_fab/datastructures/plan.py
```diff
@@ -0,0 +1,371 @@
...
+    def to_data(self):
+        return self.data
+
+    @classmethod
+    def from_data(cls, data):
+        return cls(data['start_value'])
+
+    @classmethod
+    def from_json(cls, filepath):
+        with open(filepath, 'r') as fp:
```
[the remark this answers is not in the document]
beverlylytle: For the `to_json` method, we will lose `pretty` if `compas.json_dump` is used, which is not the biggest loss.  I will note that the implementations of these methods are copied from `compas.datastructures.Datastructure`, so data structures in general don't currently support file-like objects.

src/compas_fab/datastructures/plan.py
```diff
@@ -0,0 +1,328 @@
...
+        from networkx import lexicographical_topological_sort
+        self.check_for_cycles()
+        return [self.get_action(action_id) for action_id in lexicographical_topological_sort(self.networkx)]
+
+    def get_all_linear_sorts(self):
+        """Gets all possible linear sorts respecting the dependency ids.
+
+        Returns
+        -------
+        :obj:`list` of :obj:`list of :class:`compas_fab.datastructure.Action`
```
beverlylytle, line 259:
```suggestion
        :obj:`list` of :obj:`list` of :class:`compas_fab.datastructure.Action`
```

src/compas_fab/datastructures/plan.py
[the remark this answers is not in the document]
yck011522: I think this will not work with default serialization when the object is nested within other things,

tests/datastructures/test_plan.py
```diff
@@ -0,0 +1,162 @@
...
+    other_plan = Plan.from_data(plan.data)
+    assert plan.actions.keys() == other_plan.actions.keys()
+
+    # the data attributes point to the same generator,
+    # so no testing `append_action` yet
+    plan_json = compas.json_dumps(plan.data)
+    other_plan_data = compas.json_loads(plan_json)
+    other_plan = Plan.from_data(other_plan_data)
+    assert plan.actions.keys() == other_plan.actions.keys()
+    assert other_plan.get_action(1).parameters['param'] == Frame.worldXY()
```
beverlylytle, line 91: @yck011522 Not sure what you mean about `Action.from_data` not working... This test passes.  What sort of test do you think would fail?

yck011522:
@beverlylytle 

Thanks to your explanation with example. I have a feeling you are trying to do a very specific thing in your project and tries to generalize whatever that is to a bigger class that some others can use it. 

At the moment I see the only functions for this class are to manage dependency. Which as @gonzalocasas pointed out is pretty much a compas `Graph`. Even your example with the 3 bricks can be replicated in some way with a Graph and Graph can hold arbitrary data too. In that way, I'm not sure what extra functions can users use here with your new 'Plan' class or how this class interact with other things in compas_fab such as the planning backends. Since I have mis-interpreted (maybe intentionally) that this is a path 'planning' related, maybe I would say, if this `Plan` class is designed to hold a un-Planned and Planned trajectory, then it would be useful.

I am guessing your intention for this class is to be able to have a partial plan and allow the computer to reason which things to plan first and what is next. And somehow the first plan changes the state of the scene or the robot such that the next planning depends on the result of previous plan(s). Hence the need for dependency or ordering. 

If this is the use case, maybe I share a little how I handle this case in my own code. I have a Scene class that keeps track of each object's state (their location as in transformation, their kinematic configuration if it is a RobotModel) it holds the state of the entire scene. I also have an `Action` class, I guess similar to your `Plan`. I have a dependency relationship between Actions and I can interpret that to create a single linear ordered list of Actions. Some Actions in this list require path planning, some doesn't. If they do, they can also store other path planning relation information such as AllowedCollisionMatrix. After I plan one action with a singnature similar to `PlanAction(Scene, Action) -> PlannedAction`, I can update the Scene using the PlannedAction. This for example ensures my robot's location at the end of one Action will be the same in the beginning of next Action. 

I have a feeling if this is the intention, it would be better to add these Scene and ObjectState related classed and have functions that allow the Actions/Plans and Scene states to interact. And perhaps the function to Path Plan with these Plans and States. I don't want to be discouraging and I appreciate this new PR, but I think more functions that allow these classes to interact with each other (and also with compas_fab classes) would be useful to compas_fab users.

beverlylytle:
The purpose of this data structure is to extend Graph to make it easier to create and maintain a graph that is directed and acyclic, representing a partial order on the contents of the nodes of the graph. So this structure is somewhere between a free-for-all graph and a rigid list.   Of course everything that is done with this class can be done with Graph.  Hell, every class in compas can be represented by a dictionary, but it would be super cumbersome to work always with something as primitive as a dictionary and bag of functions.  It seems like the name "Plan" is misleading, as this class has no direct link to path planning.  I liked it because it's short, ease to type and suggests something broad and generic (like "list" or "graph"). "DAG" (directed acyclic graph) also hits those notes to some degree, but is less accessible.  I am open to suggestions on the name.  

"Action" is also up for review.  It is really just a wrapper around a dictionary.  The fact that it is a class and that Python is dynamic provides for some syntactic sugar.  But it's not much more than a node in Graph, to the point that I wonder if it even should exist.

Since the intention of the class is to be nearly as generic as Graph, I wouldn't expect the Plan to know about Scenes and ObjectStates and managing interactions between them just as Graph doesn't know about Scenes or Frames or Meshes and doesn't manage interactions between them.  I think the kind of thing you are suggesting would be well accomplished by applying some variation on the Visitor Pattern to a Graph/Plan.  Each user of the Plan could design their own visitors which would allow them to have tailor-made interactions between the contents of the nodes with the contents of other nodes or with external objects like Scenes.  This kind of decoupling of data structure from function, and decoupling of data structure from other data structure, allows for much more flexible and reusable code.  I am extremely reluctant to add any feature which treats an Action has anything but a container, or which has Plan do anything but manage storing these containers and pointers between them.  

@gonzalocasas  what do you think?

beverlylytle:
I'm reading over your comment again @yck011522 and it really sounds to me like you are describing a finite state machine, which is not at all what I'm trying to accomplish with this PR.  Or do I misunderstand?

src/compas_fab/datastructures/plan.py
```diff
@@ -0,0 +1,328 @@
...
+    def get_linear_sort(self):
+        """Sorts the planned actions linearly respecting the dependency ids.
+
+        Returns
+        -------
+        :obj:`list` of :class:`compas_fab.datastructure.Action`
+        """
+        from networkx import lexicographical_topological_sort
+        self.check_for_cycles()
+        return [self.get_action(action_id) for action_id in lexicographical_topological_sort(self.networkx)]
```
beverlylytle, line 252: Do we want these to work in Ironpython?

yck011522:
@beverlylytle I guess I understand it better when you say it's a directed acyclic graph. Maybe the name Plan really gave me high hopes of what the intention of this class wants to do and have me think about automatic planning.

Then there is just two things in my mind, regarding the naming I think descriptive names are more important than sounding elegant how about call it DirectedAccylicGraph. Second thing is why is this class in compas_fab but not in compas? I guess if the intention is generic and if there is nothing specific about robotics nor fabrication, then why is it here? I think these two are interelated, can choose to solve either one, it is just that when I look at compas_fab I expect robotic fabrication related classes and functions. And since most people use the package for path planning purposes, I would expect a very high level class with a name called Plan to do what it says. Hold a trajectory or something similar. I wouldn't want to call this DAG an Action too. An action must be actionable to make sense, and your intention seems to be more generic than that.

gonzalocasas:
Ok, let's see if we can sort this out in a way that satisfies all parties ;) 
I think it's clear that `Plan` -despite being short and sweet- is not an accurate name. On the other hand, I don't like much the idea of calling it `DAG` because then it's too generic and we actually have a concrete application for it. If -eventually- we see the need of extracting a generic DAG out of this, we can do so, move that to `compas` core, and subclass it here. But let's work first on the concrete, and generalize later.

So, the core features of this new class are: 1) to specify "units of stuff that occur", 2) specify their dependencies, 3) extract those "units of stuff" in some order. There are two names that are very closely related but come from different disciplines, ones is the [partially ordered set stuff from category theory](https://en.wikipedia.org/wiki/Partially_ordered_set) and the other is [partial order plans](https://en.wikipedia.org/wiki/Partial-order_planning), and this class is somewhere in between. It lacks the solver parts for automated planning that would make it a better for for the latter option, so, I'm leaning towards `PartialOrderSet` or just `PartialOrder` (as in, this is just an ordering) (technically `poset` would also be possible, but that's cryptic imo).

Then the other important part is how to expose the traversal of the partial order. The options for traversal are probably very varied. Besides the classic algorithms for traversal, I can think of at least one that is event-based to determine triggering of dependencies (as in the `GraphScheduler` that we have somewhere else). So, I have the feeling a set of classes based on something like `PartialOrderIterator` might be a good idea, and those could contain what is currently in the lexicographical/topological/etc `get_whatever_sort`  methods.

yck011522:
The new name PartialOrder sounds great to me. Clear and descriptive. Perhaps it would be good to keep a link to the Wiki page https://en.wikipedia.org/wiki/Partial-order_planning  in the class description. Cheers

yck011522:
Regarding visitor pattern, is there no _generic visitor pattern_ we can include in the default implementation?

beverlylytle:
I'd be happy to add an `accept(visitor)` method to `PartialOrder` in a way that starts at the sources of `PartialOrder`, then propagates the visitor down to the children and then also add an abstract method `accept(visitor)` to `Action`, but I'm unsure how to make a visitor interface/template because that would probably be very specific to the project and to the different instantiations of `Action` that a user might want to use.

src/compas_fab/datastructures/partial_order.py
```diff
@@ -0,0 +1,349 @@
...
+
+class PartialOrder(Datastructure):
+    """Data structure extending :class:`compas.datastructures.Graph` for creating
+    and maintaining a partially ordered plan (directed acyclic graph).
+    The content of any event of the plan is contained in an
+    :class:`compas_fab.datastructures.Action`.  The dependency ids of a planned
+    action can be thought of as pointers to the parents of that planned action.
+    While actions can be added and removed using the methods of
+    :attr:`compas_fab.datastructures.PartialOrder.graph`, it is strongly recommended
+    that the methods ``plan_action``, ``append_action`` and ``remove_action``
```
yck011522, line 88: do you mean ``get_action`` instead of ``plan_action``?

Another thought is, how about renaming ``PartialOrder.get_action()`` to ``PartialOrder.action()``? Is there some meaning associated with the `get_` prefix?

beverlylytle:
>Also, does your implementation always visit each node one time? (Not sure if I Or maybe that is something the user has to deal with. I'm also not sure how the source_first works.

Yes, as it's written now, each node is visited exactly one.  The `source_first` flag indicates whether you want the `visitor` to do it's thing on a node and then visit the children or the other way around.  That is, does the `visitor` start at the sources of the dag or at the sinks of the dag?

I'm sure there's a way to pass a `visitor` around in a graph anyway you want, including ways that result in infinite loops.  However, the `accept` method that's in this draft, is top down or bottom up and nothing in between.  But it is a draft and I am open to suggestion.  However, it sounds like you could make do with what's in this draft by using multiple passes on your `PartialOrder` with visitors, with one doing the initial work, and another doing cleanup.

src/compas_fab/datastructures/partial_order.py
yck011522: Thanks for clarifying, I just somehow had this idea where if there is a `get_` prefix, then I should not modify the returned result in place but rather use the corresponding `set_` method. But I could be hallucinating with other language flavours.

PR #306, opened 2021-05-26
Change temp file name in convert_mesh_to_body

### LINKED ISSUE #305: PybulletClient cannot load two different collision meshes into the scene
**Describe the bug**
PybulletClient cannot load two different collision meshes into the scene. The client can only load the mesh that's loaded first.

I have the following script (full working example below):

```
with PyBulletClient(connection_type="gui", verbose=False) as client:
    base_plate_cm = parse_collision_mesh_from_path(data_dir, "base_plate.obj", scale=1)
    column_obstacle_cm = parse_collision_mesh_from_path(data_dir, "column_obstacle.obj", scale=1)
    client.add_collision_mesh(base_plate_cm)
    client.add_collision_mesh(column_obstacle_cm)
```

I expect that there are two different collision meshes in the scene:
image

But this is what `PybulletClient` gives me:
image

**To Reproduce**
Minimal working example:

```python
import os
import pybullet as p
from compas.datastructures import Mesh
from compas_fab.robots import CollisionMesh
from compas_fab.backends import PyBulletClient

HERE = os.path.dirname(__file__)

def parse_collision_mesh_from_path(dir_path, filename, scale=1e-3):
    file_path = os.path.join(dir_path, filename)
    obj_name = filename.split('.')[0]
    if filename.endswith('.obj'):
        mesh = Mesh.from_obj(file_path)
    elif filename.endswith('.stl'):
        mesh = Mesh.from_stl(file_path)
    else:
        return None
    cm = CollisionMesh(mesh, obj_name)
    cm.scale(scale)
    return cm

with PyBulletClient(connection_type="gui", verbose=False) as client:
    data_dir = HERE
    base_plate_cm = parse_collision_mesh_from_path(data_dir, "base_plate.obj", scale=1)
    column_obstacle_cm = parse_collision_mesh_from_path(data_dir, "column_obstacle.obj", scale=1)

    client.add_collision_mesh(base_plate_cm)
    client.add_collision_mesh(column_obstacle_cm)

    input()
```

[Zip file with data](https://github.com/compas-dev/compas_fab/files/6540967/pb_client_test.zip)

 - compas_fab `0.18.0`
 - Windows 10
 - Python version `3.7`
 - Python package manager `pip`

**Comments**:

In `compas_fab 0.16.0`, we used to use `tempdir` when creating the temporary mesh to be loaded by pybullet: https://github.com/compas-dev/compas_fab/blob/78efe1f87621b2bf7c83d3f8466a7e6e181553df/src/compas_fab/backends/pybullet/client.py#L429-L442

But I saw that this is removed in `0.17.0`, @gonzalocasas @beverlylytle is there a reason for this? https://github.com/compas-dev/compas_fab/blob/811b9ac8d15570d321966c38183bb9c617983d11/src/compas_fab/backends/pybullet/client.py#L625-L629

I revert the `convert_mesh_to_body` to the version used in `0.16.0` and the problem is fixed.

beverlylytle:
Closes #305   PyBullet, by default, caches the contents of obj's that it has loaded.  If it encounters a recognized path to an obj, it assumes the contents are the same as when it first read it.  Previously all the meshes were stored in `some/temp/dir/temp.obj`.  Now they will be stored in `some/temp/dir/{mesh.guid}.obj`.   I don't want to disable cacheing entirely, because that would effect performance.  I also don't want to return to having a new temp dir every time `convert_mesh_to_body` is called, because that is essentially turning off cacheing.  @gonzalocasas is this sufficient?  Given the way `mesh.uuid` is created (that is, lazily created when `guid` is called, but not modified when the mesh changes), the problem still exists for the following code:
```
from compas.datastructures import Mesh
from compas.geometry import Translation

import compas_fab
from compas_fab.backends.pybullet import PyBulletClient
from compas_fab.robots import CollisionMesh

with PyBulletClient() as client:
    mesh = Mesh.from_stl(compas_fab.get('planning_scene/floor.stl'))
    cm = CollisionMesh(mesh, 'floor')
    client.add_collision_mesh(cm)
    
    mesh.transform(Translation.from_vector([1,1,1]))
    cm = CollisionMesh(mesh, 'floor_translated')
    client.add_collision_mesh(cm)
```

### Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,8 @@ Unreleased
 
 **Fixed**
 
+* Fixed bug in ``compas.backends.PyBulletClient.convert_mesh_to_body`` circumventing PyBullet's propensity to cache
+
 **Deprecated**
 
 **Removed**
@@ -51,8 +53,8 @@ Unreleased
 
 **Changed**
 
-* The `Configuration` class has moved to `compas.robots`, but is still aliased within `compas_fab.robots`
-* Lazily load `V-REP remoteApi` library
+* The ``Configuration`` class has moved to ``compas.robots``, but is still aliased within ``compas_fab.robots``
+* Lazily load ``V-REP remoteApi`` library
 
 **Fixed**
 
@@ -64,7 +66,7 @@ Unreleased
 
 **Deprecated**
 
-* `compas_fab.robots.Configuration` is being deprecated in favor of `compas.robots.Configuration`
+* ``compas_fab.robots.Configuration`` is being deprecated in favor of ``compas.robots.Configuration``
 
 0.17.0
 ----------
--- a/src/compas_fab/backends/pybullet/client.py
+++ b/src/compas_fab/backends/pybullet/client.py
@@ -622,7 +622,7 @@ class PyBulletClient(PyBulletBase, ClientInterface):
         -------
         :obj:`int`
         """
-        tmp_obj_path = os.path.join(self._cache_dir.name, 'temp.obj')
+        tmp_obj_path = os.path.join(self._cache_dir.name, '{}.obj'.format(mesh.guid))
         mesh.to_obj(tmp_obj_path)
         tmp_obj_path = self._handle_concavity(tmp_obj_path, self._cache_dir.name, concavity, mass)
         pyb_body_id = self.body_from_obj(tmp_obj_path, concavity=concavity, mass=mass)


gonzalocasas (approved):
LGTM! When all tests are green, go for it

yijiangh:
Thanks for fixing the bug! 

> Given the way mesh.uuid is created (that is, lazily created when guid is called, but not modified when the mesh changes)

Shall we add this case as a warning in the docstring to give the users a heads-up, until we have a better solution for this?

beverlylytle:
Done.

PR #329, opened 2021-09-20
yield uniform types in inverse_kinematics

beverlylytle:
Here I've changed the ik backend features to yield the results rather than return them.

One point that is up for discussion still is that `robot.inverse_kinematics` returns a configuration while `client.inverse_kinematics` generates tuples of joint value, joint name pairs.  This is confusing.  

### Checklist

### DIFF
--- a/src/compas_fab/backends/vrep/backend_features/vrep_inverse_kinematics.py
+++ b/src/compas_fab/backends/vrep/backend_features/vrep_inverse_kinematics.py
@@ -40,9 +40,9 @@ class VrepInverseKinematics(InverseKinematics):
                   to retry infinitely.
                 - ``"max_results"``: (:obj:`int`) Maximum number of result states to return.
 
-        Returns:
-            list: List of :class:`Configuration` objects representing
-            the collision-free configuration for the ``goal_frame``.
+        Yields:
+            :obj:`tuple` of :obj:`list`
+                A tuple of 2 elements containing a list of joint positions and a list of matching joint names.
         """
         options = options or {}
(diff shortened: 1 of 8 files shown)


src/compas_fab/backends/vrep/backend_features/vrep_inverse_kinematics.py
```diff
@@ -59,5 +59,7 @@ def inverse_kinematics(self, robot, frame_WCF, start_configuration=None, group=N
 
         states = self.client.find_raw_robot_states(group, frame_to_vrep_pose(frame_WCF, self.client.scale), gantry_joint_limits, arm_joint_limits, max_trials, max_results)
 
-        return [config_from_vrep(states[i:i + num_joints], self.client.scale)
-                for i in range(0, len(states), num_joints)]
+        joint_names = robot.get_configurable_joint_names()
+
+        for i in range(0, len(states), num_joints):
+            yield config_from_vrep(states[i: i+num_joints], self.client.scale).joint_values, joint_names
```
gonzalocasas, line 65: This is cool :+1: I think now, besides the fact that Vrep has no sematics (and that messes up with the group selection), it would almost be possible to call IK from the Robot class and have it properly forwarded into V-REP.

gonzalocasas (approved):
The only bit that I wasn't sure how to handle was the `Robot.inverse_kinematics` return type, for which I sent a different PR to merge into this. Everything else LGTM!

gonzalocasas:
Linked PR: https://github.com/compas-dev/compas_fab/pull/330

PR #330, opened 2021-10-05
Make Robot.inverse_kinematics use the last generator if called repeatedly with same arguments

gonzalocasas:
PR of a PR! As discussed, this suggestion was too big to add as PR comments, so, here it is as a PR of a PR. The idea is make `inverse_kinematics` in `Robot` behave like a generator if called multiple times with same arguments, but without actually becoming a generator, just keeping track of the last one used internally.

### DIFF
--- a/src/compas_fab/robots/robot.py
+++ b/src/compas_fab/robots/robot.py
@@ -55,6 +55,10 @@ class Robot(object):
         self.semantics = semantics
         self.client = client
         self.attributes = {}
+        self._current_ik = {
+            'request_id': None,
+            'solutions': None
+        }
 
     @property
     def artist(self):
@@ -1119,6 +1123,14 @@ class Robot(object):
     def inverse_kinematics(self, frame_WCF, start_configuration=None, group=None, return_full_configuration=False, options=None):
         """Calculate the robot's inverse kinematic for a given frame.
 
+        The inverse kinematic solvers are implemented as generators which fits both analytic
+        and numerical solver approaches. However, this class abstracts that away and returns one or zero
+        configuration at a time to simplify its usage. In order to keep the usefulness of the generator,
+        calls to this method will recall the last retrieved iterator and keep returning results
+        yielded by it, one at a time.
+        This is true only as long as the request is identical to the last one. If not, the last generator
+        is discarded and the client IK implementation is called again.
+
         Parameters
         ----------
         frame_WCF : :class:`compas.geometry.Frame`
@@ -1156,6 +1168,15 @@ class Robot(object):
         >>> robot.inverse_kinematics(frame_WCF, start_configuration, group)                 # doctest: +SKIP
         Configuration((4.045, 5.130, -2.174, -6.098, -5.616, 6.283), (0, 0, 0, 0, 0, 0))    # doctest: +SKIP
         """
+        # Pseudo-memoized sequential calls will re-use iterator if not exhaused
+        request_id = '{}-{}-{}-{}-{}'.format(id(frame_WCF), id(start_configuration), id(group), id(return_full_configuration), id(options))
+
+        if self._current_ik['request_id'] == request_id and self._current_ik['solutions'] is not None:
+            ik = next(self._current_ik['solutions'], None)
+            if ik is not None:
+                joint_positions, joint_names = ik
+                return self._build_configuration(joint_positions, joint_names, group, return_full_configuration)
+
         options = options or {}
         attached_collision_meshes = options.get('attached_collision_meshes') or []
 
@@ -1172,12 +1193,20 @@ class Robot(object):
 
         options['attached_collision_meshes'] = attached_collision_meshes
 
+        solutions = self.client.inverse_kinematics(self,
+                                                   frame_WCF_scaled,
+                                                   start_configuration_scaled,
+                                                   group,
+                                                   options)
+        self._current_ik['request_id'] = request_id
+        self._current_ik['solutions'] = solutions
+
         # The returned joint names might be more than the requested ones if there are passive joints present
-        joint_positions, joint_names = next(self.client.inverse_kinematics(self,
-                                                                           frame_WCF_scaled,
-                                                                           start_configuration_scaled,
-                                                                           group,
-                                                                           options))
+        joint_positions, joint_names = next(solutions)
+
+        return self._build_configuration(joint_positions, joint_names, group, return_full_configuration)
+
+    def _build_configuration(self, joint_positions, joint_names, group, return_full_configuration):
         if return_full_configuration:
             # build configuration including passive joints, but no sorting
             joint_types = self.get_joint_types_by_names(joint_names)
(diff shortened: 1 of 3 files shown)


src/compas_fab/robots/robot.py
```diff
@@ -1156,6 +1168,15 @@ def inverse_kinematics(self, frame_WCF, start_configuration=None, group=None, re
...
         Configuration((4.045, 5.130, -2.174, -6.098, -5.616, 6.283), (0, 0, 0, 0, 0, 0))    # doctest: +SKIP
         """
+        # Pseudo-memoized sequential calls will re-use iterator if not exhaused
+        request_id = '{}-{}-{}-{}-{}'.format(id(frame_WCF), id(start_configuration), id(group), id(return_full_configuration), id(options))
+
+        if self._current_ik['request_id'] == request_id and self._current_ik['solutions'] is not None:
+            ik = next(self._current_ik['solutions'], None)
+            if ik is not None:
+                joint_positions, joint_names = ik
+                return self._build_configuration(joint_positions, joint_names, group, return_full_configuration)
```
beverlylytle, line 1178: Am I reading this correctly?  If the generator is exhausted, then `ik` is `None` and we jump over this `if` clause to then call again `inverse_kinematics`.  Is that really the intuitive behavior?  Especially in the case of analytic kinematics, I would want to know if there were only, say, 4 solutions, and not have to check if the solutions were being repeated.

beverlylytle: (This situation might come up if we implement the Descartes path planner.)

gonzalocasas: It's a good point. Indeed that is what happens (also added a unit test to verify it). From a user pov, I actually think it is good that IK keeps returning solutions and if using an analytic solver, it cycles back to start. Especially when we're talking about the default use case.  Now, for the cases in which the user actually cares about all possible solutions, this obviously won't cut it. One option would be that we add a new method to the interface, `iter_inverse_kinematics` -which btw, we had a draft PR for ;) - but this time, instead of being a new backend feature, `iter_inverse_kinematics` would simply be the same as `inverse_kinematics` but exposing the generator, so that, if the user needs to, they can get all possible solutions.

How does that sound? Should I add that?

beverlylytle: mmm... i dunno.  i guess my issue is the different behaviors of the various iks.  `list(robot.inverse_kinematics(...))` causes an infinite loop, but (with your suggestion) `list(client.inverse_kinematics(...))` and `list(robot.iter_inverse_kinematics(...))` wouldn't.  not the worst thing, but also not intuitive/consistent.

beverlylytle: oh, i retract the previous comment.  `robot.inverse_kinematics` returns things, so wouldn't create an infinite loop, but i guess it's still a similar sort of issue.  the robot's ik has a different return type than the client's...

gonzalocasas: > the robot's ik has a different return type than the client's

yes, that is a bit odd at first, but I guess we both considered that the right thing to do is to keep `Robot.inverse_kinematics`' returning values instead of being a generator. Changing it to a generator would impact a lot of code, and it would make even the simplest use cases more complex than needed. 

So, the main issue that remains is the fact that a same-named method returns different things, right? However, these two methods are already quite different from each other (besides the yield vs return): one outputs configurations, the other joint positions+joint names tuples.

If we change the backend features to have one for `iter_inverse_kinematics` and another for `inverse_kinematics` we would be replicating the situation we had in the draft PR, which was not very convincing, and it adds the need to have extra code in all implementations to handle both. With the proposal of adding `Robot.iter_inverse_kinematics` only, and keeping the rest as is, we would automatically get both behaviors (iterable and single-shot) for all implementations of IK.

beverlylytle: yeah, `iter_inverse_kinematics` it is then.

src/compas_fab/robots/robot.py
```diff
@@ -1156,6 +1168,15 @@ def inverse_kinematics(self, frame_WCF, start_configuration=None, group=None, re
         >>> robot.inverse_kinematics(frame_WCF, start_configuration, group)                 # doctest: +SKIP
         Configuration((4.045, 5.130, -2.174, -6.098, -5.616, 6.283), (0, 0, 0, 0, 0, 0))    # doctest: +SKIP
         """
+        # Pseudo-memoized sequential calls will re-use iterator if not exhaused
+        request_id = '{}-{}-{}-{}-{}'.format(id(frame_WCF), id(start_configuration), id(group), id(return_full_configuration), id(options))
```
beverlylytle, line 1172: actually, i was just looking into `id` and the docs say "This is an integer which is guaranteed to be unique and constant for this object during its lifetime." So if an attribute of the frame changes or the value of some pair in the options change, then the `id` doesn't change.   Yay mutability! Do we want that?

beverlylytle: oh, and this is probably a one in a million bug, but 'Two objects with non-overlapping lifetimes may have the same id() value.'

gonzalocasas: Oh my! good catch! especially the mutability part of the dict. I'll go for `str()` instead of `id()`

src/compas_fab/backends/pybullet/backend_features/pybullet_inverse_kinematics.py
```diff
@@ -76,57 +88,111 @@ def inverse_kinematics(self, robot, frame_WCF, start_configuration=None, group=N
         if start_configuration:
             start_configuration = self.client.set_robot_configuration(robot, start_configuration, group)
```
beverlylytle, line 89: this doesn't look right.  shouldn't it be set to the zero config if `start_configuration` isn't given?

gonzalocasas: yes, i think that happens in line 92

beverlylytle: but the configuration is not `set` in the pybullet server.  so it uses the current pose as the seed rather than the zero config as suggested by the doc string (considering it seems that `restPoses` doesn't really have an effect).

src/compas_fab/backends/pybullet/backend_features/pybullet_inverse_kinematics.py
```diff
@@ -145,11 +146,12 @@ def inverse_kinematics(self, robot, frame_WCF, start_configuration=None, group=N
 
                 # ik_options.update(dict(
                 #     residualThreshold=options.get('high_accuracy_threshold', 1e-6),
-                #     maxNumIterations=options.get('high_accuracy_max_iter', 300),
+                #     maxNumIterations=options.get('high_accuracy_max_iter', 20),
                 # ))
                 # joint_positions = pybullet.calculateInverseKinematics(**ik_options)
             else:
                 joint_positions = pybullet.calculateInverseKinematics(**ik_options)
+                close_enough = True     # yeah, let's say we trust it
```
beverlylytle, line 154: 😂

PR #359, opened 2022-06-08
Gh const from config

chenkasirer:
Created a `ConstraintsFromConfiguration` GH component. Generated a new icon for `ConstraintsFromPlane` component to fit.

### Checklist


src/compas_fab/ghpython/components/Cf_ConstraintsFromConfiguration/code.py
```diff
@@ -0,0 +1,24 @@
...
+import math
+
+from ghpythonlib.componentbase import executingcomponent as component
+
+
+class ConstraintsFromTargetConfiguration(component):
+    def RunScript(self, robot, target_configuration, tolerance_above, tolerance_below, group_name):
+        if robot and target_configuration:
+            tolerance_above = tolerance_above or [math.radians(1)] * 6
+            tolerance_below = tolerance_below or [math.radians(1)] * 6
```
gonzalocasas, line 15: This one is a bit more involved to be generic: it won't always be 6 values and it won't always be in radians. So, you need to get a `get_configurable_joints(group)` for the robot, and add defaults based on the type of each joint (for PRIMASTIC and PLANAR, the unit is meters, for all the rest, radians).

As an example of a more annoying case, the SCARA robots have 4 axes (revolute, revolute, prismatic, revolute), so, it should end up in `radians(1), radians(1), 0.001, radians(1)` (the prismatic could default to 1mm, but it is defined in meters).

gonzalocasas (changes requested):
Only one comment that should be addressed before merging, but otherwise LGTM!

src/compas_fab/ghpython/components/Cf_ConstraintsFromConfiguration/code.py
chenkasirer: Good point. 
Is there maybe a more generic attribute I should check? (e.g. `is_scalable()`) or is this not directly related to the tolerance unit.

gonzalocasas: Yes, i think you can rely on `is_scalable` because angles cannot be scaled so this has matching semantics.

src/compas_fab/ghpython/components/Cf_ConstraintsFromConfiguration/code.py
```diff
@@ -0,0 +1,35 @@
...
+
+from ghpythonlib.componentbase import executingcomponent as component
+
+
+class ConstraintsFromTargetConfiguration(component):
+
+    DEFAULT_TOLERANCE_METERS = .001
+    DEFAULT_TOLERANCE_RADIANS = math.radians(1)
+
+    def RunScript(self, robot, target_configuration, tolerance_above, tolerance_below, group_name):
```
gonzalocasas, line 16:
```suggestion
    def RunScript(self, robot, target_configuration, tolerance_above, tolerance_below, group):
```
I would rename this to `group` to match all other components. I guess this works because GH matches the input arguments on the metadata to arguments of this function purely based on position.

gonzalocasas (approved):
LGTM! Only one tiny suggestion

PR #365, opened 2022-08-18
Planning scene viz

gonzalocasas:
This change (finally) adds supports for all the planning scene elements to be visualized with the default `Robot Visualize` component of Grasshopper. This means both attached and non-attached collision meshes will be retrieved and shown and there's no need to pass them to the component explicitly anymore.

Here's a quick screenshot showing the component in action with both an ACM and a CM:

image
image

It also changes the behavior of the `Attach tool` component that was (needlessly) adding the tool meshes to the planning scene, but they can be an in-memory thing only.

And a bunch of typos as well.

### Checklist

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -12,7 +12,11 @@ Unreleased
 
 **Added**
 
+* Added support for attached and non-attached collision mesh visualization to the ``Robot Visualize`` GH component.
+
 **Changed**
+
+* Changed behavior of ``Attach Tool`` GH component to only attach the tool but not add it to the planning scene state.
 
 **Fixed**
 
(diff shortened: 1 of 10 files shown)


gonzalocasas:
After updating `flake8`, there were a bunch of additional lint errors, so, the last commit is a `black` formatting of two files to fix those.

chenkasirer (approved):
LGTM!

So this was probably already outdated before, but maybe worth updating [this](https://gramaziokohler.github.io/compas_fab/latest/examples/03_backends_ros/06_ros_in_grasshopper.html#grasshopper-components) accordingly at some point :)

Also, and this is general, I think we should either embrace the standard `black` line length, or add a `pyproject.toml` with the COMPAS 180 char length configuration cause it's kinda annoying to have to add `-l 180` everytime, and is very easy to forget.

gonzalocasas:
> add a `pyproject.toml` with

+1 to that!

PR #368, opened 2022-08-19
Small updates to collision objects in ROS

gonzalocasas:
Mostly just a doc update to collision object management in ROS, but additionally, this adds the `append` operation to the `Collision Mesh` GH component and fixes a mistake (that I introduced yesterday) in the `Visualize Robot` component, because `aco.object` is not an instance of a ros message class, it's only a dictionary (`frame_id = aco.object["header"]["frame_id"]`)

Now the docs look like this:
image
image
image
image
image

- [x] Other (doc update)

### Checklist


src/compas_fab/ghpython/components/Cf_CollisionMesh/code.py
```diff
@@ -10,14 +10,17 @@
...
-    def RunScript(self, scene, M, name, add, remove):
+    def RunScript(self, scene, M, name, add, append, remove):
         ok = False
         if scene and M and name:
             mesh = RhinoMesh.from_geometry(M).to_compas()
             collision_mesh = CollisionMesh(mesh, name)
             if add:
                 scene.add_collision_mesh(collision_mesh)
                 ok = True
+            if append:
```
chenkasirer, line 21: Does it make sense to both add and append a mesh to the scene?

gonzalocasas: I think it makes sense on the same component. Obviously not for the same object, but if you are trying to add multiple similar meshes (such as the brick in the GH example i added), then it makes sense to have the option of appending instead of normal add. 

One thing though is that "add/append" have here very explicit ROS-related meaning. Perhaps should we "hide" this behind a flag like "remove before adding"?

chenkasirer: Makes perfect sense with the example you gave. I was just wondering what would happen if you plug True boolean flags to both arguments at the same time. Since add is checked first, I guess the mesh will get replaced?

Maybe the names make perfect sense for people who are familiar with the workflow, I was thinking more if we maybe need to check for mutual exclusivity between the flags. Dunno if we usually make checks like that in components.

chenkasirer (approved):
LGTM!

src/compas_fab/ghpython/components/Cf_CollisionMesh/code.py
gonzalocasas: Added some sanity check for mutual exclusivity!

PR #370, opened 2022-08-19
Moving most config to pyproject.toml

gonzalocasas:
First added the max line length to `black` settings in a new `pyproject.toml` and then moved most of the settings from `setup.cfg` to `pyproject.toml`. I tried to move `bdist_wheel` setting but it didn't seem to use it even though it is technically supported (I might have missed another setting to trigger that?), and then `flake8` would be the other one remaining, which doesn't officially support `pyproject.toml`, but we should replace with `black`.

### Checklist

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,7 +7,7 @@ check-manifest>=0.36
 flake8
 autopep8
 pylint
-pytest>=4.6
+pytest>=6.0
 pytest_mock
 pytest-cov
 sybil
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 attrs>=19.3.0
-autopep8
+black
 bump2version>=1.0
 check-manifest>=0.36
 flake8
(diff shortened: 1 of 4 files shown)


requirements-dev.txt
```diff
@@ -1,16 +1,16 @@
 attrs>=19.3.0
-sphinx_compas_theme>=0.9
-sphinx>=1.6
-invoke>=0.14
+autopep8
```
chenkasirer, line 2: do we use this anywhere? should we not just replace with `black`?

gonzalocasas: We haven't officially made the switch to black yet. I didn't want to mix the switch to black with this PR, i can send an immediate follow up with switch plus a general run of black over the entire code.

chenkasirer: Ah totally at your discretion. I view the move to black as merely changing from `autopep8` to `black` in `reqirements.txt` . Dunno if I'd run `black` on everything at once. Would be a frightening PR to review ;)

chenkasirer (approved):
LGTM

requirements-dev.txt
gonzalocasas: Ok, then let's do the switch to `black` in this PR, and I will check how scary a full run of black is for a single PR.

PR #381, opened 2023-04-23
Add scene cache to 'Visualize Robot" component

gonzalocasas:
This is a small but noticeable performance optimization when there are lots of collision meshes in the scene. It prevents the retrieval and reloading of the scene constantly, and instead caches it for a couple of seconds. 
This very short cache is enough to prevent reloading while the user is, eg. dragging a slider to update the robot:
perf-viz-robot

### Checklist

### DIFF
--- a/src/compas_fab/ghpython/components/Cf_VisualizeRobot/code.py
+++ b/src/compas_fab/ghpython/components/Cf_VisualizeRobot/code.py
@@ -3,11 +3,16 @@ Visualizes the robot.
 
 COMPAS FAB v0.27.0
 """
+import time
+
 from compas.artists import Artist
 from compas.geometry import Frame
 from compas.geometry import Transformation
-from compas_fab.robots import PlanningScene
 from ghpythonlib.componentbase import executingcomponent as component
+from scriptcontext import sticky as st
+
+from compas_fab.ghpython.components import create_id
+from compas_fab.robots import PlanningScene
 
 
 class RobotVisualize(component):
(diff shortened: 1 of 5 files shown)


src/compas_fab/ghpython/components/Cf_VisualizeRobot/code.py
```diff
@@ -3,11 +3,16 @@
...
+import time
+
 from compas.artists import Artist
 from compas.geometry import Frame
 from compas.geometry import Transformation
-from compas_fab.robots import PlanningScene
 from ghpythonlib.componentbase import executingcomponent as component
+from scriptcontext import sticky as st
+
+from compas_fab.ghpython.components import create_id
```
chenkasirer, line 14: Maybe replace with `create_id` from `compas_ghpython`? one step towards retiring this one

chenkasirer (approved):
lgtm

src/compas_fab/ghpython/components/Cf_VisualizeRobot/code.py
gonzalocasas: it would be nice, but it's still not released and it would bump the minimum requirement of compas for something rather trivial. What I could do thou, is `try..except` the import and keep the current function as fallback if it's not available.. not sure if it's worth it though

chenkasirer: ah that's a good point. nevermind then.

PR #387, opened 2023-10-04
Changed classes to inherete from compas.Data

yck011522:
Some of the classes in compas_fab is still based on (object) instead of (Data)

I have selected some of the classes that should support serialization and upgrade them. Upgrading them should be backward compatible. However, it may affect user classes inherited from these classes. This RP does not cover all the classes yet (e.g. classes in backends, reachability map, ghpython and some more). Let me know if you think I should extend to cover all classes. 

@gonzalocasas Should I add test files for this? 

### Checklist

### DIFF
--- a/src/compas_fab/robots/robot.py
+++ b/src/compas_fab/robots/robot.py
@@ -4,6 +4,7 @@ from __future__ import print_function
 
 import random
 
+from compas.data import Data
 from compas.geometry import Frame
 from compas.geometry import Sphere
 from compas.geometry import Transformation
@@ -21,7 +22,7 @@ __all__ = [
 ]
 
 
-class Robot(object):
+class Robot(Data):
     """Represents a robot.
 
     This class binds together several building blocks, such as the robot's
@@ -47,15 +48,44 @@ class Robot(object):
         Dictionary mapping planning groups to the tool currently attached to them, if any.
     """
 
-    def __init__(self, model, artist=None, semantics=None, client=None):
+    def __init__(self, model=None, artist=None, semantics=None, client=None):
+        super(Robot, self).__init__()
+        # These attributes have to be initiated first, 
+        # because they are used in the setters of the other attributes
         self._scale_factor = 1.0
-        self.model = model
         self._attached_tools = {}  # { planning_group_name: robots.tool.Tool }
+        self._current_ik = {"request_id": None, "solutions": None}
+
+        self.model = model
         self.artist = artist
         self.semantics = semantics
         self.client = client
         self.attributes = {}
-        self._current_ik = {"request_id": None, "solutions": None}
+
+    @property
+    def data(self):
+        data = {
+            '_scale_factor': self._scale_factor,
+            '_attached_tools': self._attached_tools,
+            '_current_ik': self._current_ik,
+            'model': self.model.data,
+            # 'artist': self.artist.data if self.artist else None,
+            # 'semantics': self.semantics.data if self.semantics else None,
+            # 'client': self.client.data if self.client else None,
+            'attributes': self.attributes,
+        }
+        return data
+
+    @data.setter
+    def data(self, data):
+        self._scale_factor = data.get('_scale_factor', 1.0)
+        self._attached_tools = data.get('_attached_tools', {})
+        self._current_ik = data.get('_current_ik', {"request_id": None, "solutions": None})
+        self.model = RobotModel.from_data(data['model'])
+        # self.artist = data.get('artist', None)
+        # self.semantics = data.get('semantics', None)
+        # self.client = data.get('client', None)
+        self.attributes = data.get('attributes', {})
 
     @property
     def artist(self):
@@ -65,6 +95,8 @@ class Robot(object):
     @artist.setter
     def artist(self, artist):
         self._artist = artist
+        if artist is None:
+            return
         if len(self.model.joints) > 0 and len(self.model.links) > 0:
             self.scale(self._scale_factor)
             for tool in self.attached_tools.values():
--- a/src/compas_fab/robots/robot.py
+++ b/src/compas_fab/robots/robot.py
@@ -65,9 +65,10 @@ class Robot(Data):
     @property
     def data(self):
         data = {
-            "_scale_factor": self._scale_factor,
-            "_attached_tools": self._attached_tools,
-            "_current_ik": self._current_ik,
+            "scale_factor": self._scale_factor,
+            "attached_tools": self._attached_tools,
+            # The current_ik is an extrinsic state that is not serialized with the robot
+            # "current_ik": self._current_ik,
             "model": self.model.data,
             "semantics": self.semantics,
             "attributes": self.attributes,
@@ -79,9 +80,8 @@ class Robot(Data):
 
     @data.setter
     def data(self, data):
-        self._scale_factor = data.get("_scale_factor", 1.0)
-        self._attached_tools = data.get("_attached_tools", {})
-        self._current_ik = data.get("_current_ik", {"request_id": None, "solutions": None})
+        self._scale_factor = data.get("scale_factor", 1.0)
+        self._attached_tools = data.get("attached_tools", {})
         self.model = RobotModel.from_data(data["model"])
         self.semantics = data.get("semantics", None)
         self.attributes = data.get("attributes", {})
--- a/src/compas_fab/robots/robot.py
+++ b/src/compas_fab/robots/robot.py
@@ -69,10 +69,10 @@ class Robot(Data):
             "_attached_tools": self._attached_tools,
             "_current_ik": self._current_ik,
             "model": self.model.data,
+            "semantics": self.semantics,
             "attributes": self.attributes,
             # The following attributes cannnot be serizlied
             # "artist": self.artist.data if self.artist else None,
-            # "semantics": self.semantics.data if self.semantics else None,
             # "client": self.client.data if self.client else None,
         }
         return data
@@ -83,6 +83,7 @@ class Robot(Data):
         self._attached_tools = data.get("_attached_tools", {})
         self._current_ik = data.get("_current_ik", {"request_id": None, "solutions": None})
         self.model = RobotModel.from_data(data["model"])
+        self.semantics = data.get("semantics", None)
         self.attributes = data.get("attributes", {})
 
     @property
--- a/src/compas_fab/robots/robot.py
+++ b/src/compas_fab/robots/robot.py
@@ -50,7 +50,7 @@ class Robot(Data):
 
     def __init__(self, model=None, artist=None, semantics=None, client=None):
         super(Robot, self).__init__()
-        # These attributes have to be initiated first, 
+        # These attributes have to be initiated first,
         # because they are used in the setters of the other attributes
         self._scale_factor = 1.0
         self._attached_tools = {}  # { planning_group_name: robots.tool.Tool }
@@ -65,27 +65,25 @@ class Robot(Data):
     @property
     def data(self):
         data = {
-            '_scale_factor': self._scale_factor,
-            '_attached_tools': self._attached_tools,
-            '_current_ik': self._current_ik,
-            'model': self.model.data,
-            # 'artist': self.artist.data if self.artist else None,
-            # 'semantics': self.semantics.data if self.semantics else None,
-            # 'client': self.client.data if self.client else None,
-            'attributes': self.attributes,
+            "_scale_factor": self._scale_factor,
+            "_attached_tools": self._attached_tools,
+            "_current_ik": self._current_ik,
+            "model": self.model.data,
+            "attributes": self.attributes,
+            # The following attributes cannnot be serizlied
+            # "artist": self.artist.data if self.artist else None,
+            # "semantics": self.semantics.data if self.semantics else None,
+            # "client": self.client.data if self.client else None,
         }
         return data
 
     @data.setter
     def data(self, data):
-        self._scale_factor = data.get('_scale_factor', 1.0)
-        self._attached_tools = data.get('_attached_tools', {})
-        self._current_ik = data.get('_current_ik', {"request_id": None, "solutions": None})
-        self.model = RobotModel.from_data(data['model'])
-        # self.artist = data.get('artist', None)
-        # self.semantics = data.get('semantics', None)
-        # self.client = data.get('client', None)
-        self.attributes = data.get('attributes', {})
+        self._scale_factor = data.get("_scale_factor", 1.0)
+        self._attached_tools = data.get("_attached_tools", {})
+        self._current_ik = data.get("_current_ik", {"request_id": None, "solutions": None})
+        self.model = RobotModel.from_data(data["model"])
+        self.attributes = data.get("attributes", {})
 
     @property
     def artist(self):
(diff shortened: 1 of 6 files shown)


gonzalocasas:
> Should I add test files for this?

yes please! they don't need to be super elaborate, but testing the serialization and deserialization on all these classes with some basic data would help ensure they behave properly all the time

yck011522:
I have revised the list of affected classes:
- CollisionMesh (using existing `to_data()` `from_data()` )
- AttachedCollisionMesh (using existing `to_data()` `from_data()` )
- Robot (using existing data getter setter)
- RobotSemantics (added new data getter setter and `to_data()`)
- Tool (using existing data getter setter)

The following class is skipped as it maybe redesigned in near future
- PlanningScene

yck011522:
ping @gonzalocasas

src/compas_fab/robots/robot.py
```diff
@@ -47,15 +48,43 @@ class Robot(object):
...
+            # "artist": self.artist.data if self.artist else None,
+            # "client": self.client.data if self.client else None,
+        }
+        return data
+
+    @data.setter
+    def data(self, data):
+        self._scale_factor = data.get("_scale_factor", 1.0)
+        self._attached_tools = data.get("_attached_tools", {})
+        self._current_ik = data.get("_current_ik", {"request_id": None, "solutions": None})
```
chenkasirer, line 84: mmh is this to stay compatible with previously serialized `Robot`s? But it didn't implement the `Data` interface before, so there shouldn't be any, right?

You could also simply omit these from the `Data` dictionary if they take default values, `__init__` is ran first anyways. Otherwise you can assume they're there, because they're put there by the getter.

src/compas_fab/robots/robot.py
```diff
@@ -47,15 +48,43 @@ class Robot(object):
...
         self.client = client
         self.attributes = {}
-        self._current_ik = {"request_id": None, "solutions": None}
+
+    @property
+    def data(self):
+        data = {
+            "_scale_factor": self._scale_factor,
+            "_attached_tools": self._attached_tools,
+            "_current_ik": self._current_ik,
```
chenkasirer, line 70: why the `_`s?

src/compas_fab/robots/semantics.py
```diff
@@ -41,6 +43,37 @@ def __init__(
...
+        self.disabled_collisions = data.get("disabled_collisions", set())
+        if len(self.disabled_collisions) > 0:
+            self.disabled_collisions = {tuple(pair) for pair in self.disabled_collisions}
+        self.group_states = data.get("group_states", {})
+
+    @classmethod
+    def from_data(cls, data):
+        robot_semantics = cls(None)
+        robot_semantics.data = data
+        return robot_semantics
```
chenkasirer, line 75: @gonzalocasas commented on this. I'd say setting `robot_model=None` in arguments of `__init__` is legit. Then you don't need a custom `from_data`. While I was not a fan, this convention grew on me.

Having said that, the trend in COMPAS 2.0 is to get rid of the `data.setter` acknowledging it's often redundant. So you'd then just have a `data` getter and a `from_data` which feeds the `data` dict to `__init__`.

Example (just for illustration, not sure if this works in in LTS): https://github.com/gramaziokohler/compas_timber/commit/d11d83e268a679b016ad60ef9f5be1f027887b48#diff-5cbdc4545536c895404b80de4fdcc522ec5b95dfb4af73a47e4120888cc22031R42-R48

tests/robots/test_robot.py
```diff
@@ -3,6 +3,7 @@
 import re
 
 import pytest
+from compas.data import json_dumps, json_loads
```
chenkasirer, line 6: not so important cause unittest, but generally one import per line

tests/robots/test_robot.py
```diff
@@ -289,6 +290,51 @@ def test_get_configurable_joints_wo_semantics(panda_robot_instance_wo_semantics)
...
+    for tool in robot2.attached_tools.values():
+        assert isinstance(tool, Tool)
+    assert len(robot2.attached_tools) == 1
+    assert robot.main_group_name == robot2.main_group_name
+    tool1 = robot.attached_tool
+    tool2 = robot2.attached_tool
+    assert tool1.name == tool2.name
+    assert str(tool1.frame) == str(tool2.frame)
+
+
```
chenkasirer, line 337: I'd add another unittest checking deepcopy. should behave the same as (de)serializing but sometimes doesn't

chenkasirer (approved):
LGTM. left some comments though.

src/compas_fab/robots/robot.py
yck011522: I didn't invent this particular variable, and it was there all along, I just serialized it. I believe the intent of the existing code is that the model is not supposed to keep track of extrinsic or modifiable properties of the robot (in this case, the `current_ik` ). Hence the private `_` notation. I should probably have another PR to address the intrinsic vs extrinsic state of the robot problem.

src/compas_fab/robots/semantics.py
yck011522: I'm not sure if I get your argument. 

Are you saying that the style is moving towards having default values for all parameters in the `__init__` and then skipping the `def from_data()`?

Probably it would be good to have a more wholistic stylistic clean up separately. The style currently used in compas_fab is a mixed bag of things. I can do that as long as we can have some consensus on what convention we adopt.

tests/robots/test_robot.py
yck011522: Fixed

tests/robots/test_robot.py
yck011522: I tried deepcopy but It doesn't seem to work. I think some years ago, I heard from someone to not use deepcopy on these objects. I don't remember why. But basically I never deepcopied any compas objects since.

Just to give you a taste of that error message:

```
./tests/robots/test_robot.py::test_robot_deepcopy_with_tool Failed: [undefined]AttributeError: 'Visual' object has no attribute '_guid'
tests\robots\test_robot.py:345: in test_robot_deepcopy_with_tool
    robot2 = deepcopy(robot)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:172: in deepcopy
    y = _reconstruct(x, memo, *rv)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:271: in _reconstruct
    state = deepcopy(state, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:146: in deepcopy
    y = copier(x, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:146: in deepcopy
    y = copier(x, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:146: in deepcopy
    y = copier(x, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:172: in deepcopy
    y = _reconstruct(x, memo, *rv)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:271: in _reconstruct
    state = deepcopy(state, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:146: in deepcopy
    y = copier(x, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:146: in deepcopy
    y = copier(x, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:172: in deepcopy
    y = _reconstruct(x, memo, *rv)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:271: in _reconstruct
    state = deepcopy(state, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:146: in deepcopy
    y = copier(x, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:146: in deepcopy
    y = copier(x, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:146: in deepcopy
    y = copier(x, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:206: in _deepcopy_list
    append(deepcopy(a, memo))
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:172: in deepcopy
    y = _reconstruct(x, memo, *rv)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:271: in _reconstruct
    state = deepcopy(state, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:146: in deepcopy
    y = copier(x, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:146: in deepcopy
    y = copier(x, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:231: in _deepcopy_dict
    y[deepcopy(key, memo)] = deepcopy(value, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:146: in deepcopy
    y = copier(x, memo)
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:206: in _deepcopy_list
    append(deepcopy(a, memo))
..\..\..\anaconda3\envs\cf-dev\Lib\copy.py:161: in deepcopy
    rv = reductor(4)
..\..\..\anaconda3\envs\cf-dev\Lib\site-packages\compas\data\data.py:121: in __getstate__
    "guid": str(self.guid),
..\..\..\anaconda3\envs\cf-dev\Lib\site-packages\compas\data\data.py:200: in guid
    if not self._guid:
E   AttributeError: 'Visual' object has no attribute '_guid'
```

yck011522:
ping @gonzalocasas

src/compas_fab/robots/robot.py
gonzalocasas: I would remove the `_` in the keys.

gonzalocasas: also, `current_ik` should not be serialized

src/compas_fab/robots/semantics.py
```diff
@@ -41,6 +43,37 @@ def __init__(
...
 
+    @property
+    def data(self):
+        data = {
+            "robot_model": self.robot_model,
+            "groups": self.groups,
+            "main_group_name": self.main_group_name,
+            "passive_joints": self.passive_joints,
+            "end_effectors": self.end_effectors,
+            "disabled_collisions": sorted(self.disabled_collisions),
```
gonzalocasas, line 54: Why are these sorted? not that is bad, just curious

tests/robots/test_robot.py
```diff
@@ -289,6 +291,52 @@ def test_get_configurable_joints_wo_semantics(panda_robot_instance_wo_semantics)
...
 
+def test_semantics_serialization(panda_srdf, panda_urdf):
+    model = RobotModel.from_urdf_file(panda_urdf)
+    semantics = RobotSemantics.from_srdf_file(panda_srdf, model)
+    semantics_string = json_dumps(semantics)
+    semantics2 = json_loads(semantics_string)
+    assert isinstance(semantics, RobotSemantics)
+    assert isinstance(semantics2, RobotSemantics)
+    for disabled_collision in semantics.disabled_collisions:
+        assert disabled_collision in semantics2.disabled_collisions
```
gonzalocasas, line 302: here you should assert also main group name and maybe passive joints as also.
Also, I'm surprised the membership check (`in`) works,

src/compas_fab/robots/robot.py
```diff
@@ -47,15 +48,43 @@ class Robot(object):
...
+            "model": self.model.data,
+            "semantics": self.semantics,
+            "attributes": self.attributes,
+            # The following attributes cannnot be serizlied
+            # "artist": self.artist.data if self.artist else None,
+            # "client": self.client.data if self.client else None,
```
gonzalocasas, line 76:
```suggestion
            # The following attributes cannot be serialized: artist, client
```

gonzalocasas (commented):
Generally LGTM, I added some feedback though

tests/robots/test_robot.py
chenkasirer: it's known `deepcopy` errors produce the most beautiful stacks.. 
To this specific one, could there be a class called `Visual` (really?) somewhere which inherits from `Data` but doesn't call `super().__init__()`?

src/compas_fab/robots/semantics.py
chenkasirer: yea sorry didn't wanna cause confusion. Let's leave it like that.

src/compas_fab/robots/semantics.py
yck011522: Okay, I removed the underscore in the keys of the serialized dictionary. But kept the original variable names with underscore.

Skipped `current_ik ` serialization.

The sorted thing is to improve the diff-ability of the text output. Otherwise random ordering on save can create phantom changes that show up in diff, but in fact there are no real changes to the data.

tests/robots/test_robot.py
yck011522: Main group name assertion added.

I think the membership check is passing the string key around. So that works with checking the key in other_dict.

tests/robots/test_robot.py
yck011522: https://github.com/compas-dev/compas/blob/ba932fcb55b2839ef493f4181b1f080f377a772b/src/compas/robots/model/link.py#L166-L191

It looks okay here. I'm not sure how this Visual object is created. It seems that there is a Geometry class within it and that is created during `from_data()`:

https://github.com/compas-dev/compas/blob/ba932fcb55b2839ef493f4181b1f080f377a772b/src/compas/robots/model/link.py#L244-L248

yck011522: If someone wants to debug this, this is the test code:

```
def test_robot_deepcopy(panda_robot_instance):
    import copy
    robot = panda_robot_instance
    robot.scale(0.001)
    robot2 = copy.deepcopy(robot)
    assert robot is not robot2

def test_robot_deepcopy2(ur5_robot_instance, robot_tool1):
    import copy
    robot = ur5_robot_instance
    robot.attach_tool(robot_tool1)
    robot2 = copy.deepcopy(robot)
    assert robot is not robot2
```

Both test won't work and they have different error message

PR #389, opened 2023-10-04
fix CI dependency on compas

gonzalocasas:
The CI workflow for IronPython depends on Debian's mirror of PyPi to install wheels, such as compas, because pip generally fails miserably, but now that compas 2.0alpha is released, the debian mirror incorrectly takes that version as `/latest` and there's no `/stable` redirect, so, we need to pin to the latest LTS version.

Once `compas_fab` moves to 2.x, we could unpin this again.

### Checklist

- [ ] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,7 @@ jobs:
       - uses: actions/checkout@v2
       - name: Install dependencies
         run: |
-          curl -o compas.tar.gz -LJO https://pypi.debian.net/compas/latest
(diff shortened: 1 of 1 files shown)


yck011522 (commented):
LGTM. I guess we cannot do semantic versioning here?

gonzalocasas:
> I guess we cannot do semantic versioning here?

Unfortunately this service (the debian mirror doesn't seem to support it. I tried different variations but nothing worked.

PR #397, opened 2024-01-24
Targets Classes for plan_motion and plan_cartesian_motion

yck011522:
Creating Targets Classed for the lowest level planning API to consume. 

Some follow up PR will modify the plan_motion and plan_cartesian_motion API to take this new class type. The goal of this is to have a slightly higher level of abstract (Frames / Config) instead of the current practice of converting the targets to list of Constraints for the planning call.

This will allow the PyBullet planner, which does not support Constraints to understand the Frame and Config targets.

### Checklist

- [ ] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).


yck011522:
I made a mess. Will do the branch again.

PR #408, opened 2024-02-21
Raise specific exceptions if backend feature is not supported

gonzalocasas:
Without a specific exception, the "visualize robot" component cannot know why the planning scene part is failing when using the analytical IK solvers. So, this adds a new exception type and the component now checks for it and does not fail in that case, it simply disables showing things that will not work.

### Checklist

### DIFF
--- a/src/compas_fab/ghpython/components/Cf_VisualizeRobot/code.py
+++ b/src/compas_fab/ghpython/components/Cf_VisualizeRobot/code.py
@@ -14,6 +14,7 @@ from compas_rhino.conversions import frame_to_rhino_plane
 from ghpythonlib.componentbase import executingcomponent as component
 from scriptcontext import sticky as st
 
+from compas_fab.backends import BackendFeatureNotSupportedError
 from compas_fab.robots import PlanningScene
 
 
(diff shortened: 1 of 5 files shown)


src/compas_fab/ghpython/components/Cf_VisualizeRobot/code.py
```diff
@@ -89,7 +90,13 @@ def RunScript(
 
                 if update_scene:
                     scene = PlanningScene(robot)
-                    scene = robot.client.get_planning_scene()
+                    try:
+                        scene = robot.client.get_planning_scene()
+                    except BackendFeatureNotSupportedError:
+                        print("The selected backend does not support displaying collision meshes. Feature disabled.")
```
yck011522, line 96: I think helpful error messages should follow a pattern that is roughly:

- Explain clearly that there is a problem
- Explain what the problem is
- If possible, provide a solution so that the user can complete the process, or
- Point them to where they can go for help
- Make a frustrating scenario as pleasant as possible

In this case, I am not sure what you mean by "Feature disabled." Which feature? Is the whole Visualize Robot component being disabled? Is the collision meshes being disabled? (aka. Do I still have the CMs in the backend?) 
And then what I can do about this? How can I visualize the CMs? Maybe the backend doesn't support CMs in the first place and there are nothing to show. Maybe the backend has CMs, and we just cannot get them out. Maybe I should try another backend that supports the feature.

yck011522 (commented):
lgtm with a suggestion in error message wording.

yck011522 (approved):

src/compas_fab/ghpython/components/Cf_VisualizeRobot/code.py
gonzalocasas: good point 👍  fixed at

PR #417, opened 2024-04-19
Fix Addition error in Duration() 

yck011522:
Changed the behavior of Duration class when accepting both seconds (float) and nanoseconds (int) where the decimal point of seconds and the nanoseconds add up to more than 1 second.

In the following test (existing), the total duration is expected to be 3.1 seconds. 
```
def test_sec_remainder_add_to_nsec():
    d = Duration(2.6, 5e8)
    assert d.seconds == 3.1
```
However, existing implementation did not carry on the values from nsec to sec when the total value adds up.  The following test explains the more desirable behavior:

```
def test_sec_remainder_add_to_nsec():
    d = Duration(2.6, 5e8)
    assert d.secs == 3
    assert d.nsecs == 1e8
    assert d.seconds == 3.1
```

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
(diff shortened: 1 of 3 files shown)


gonzalocasas (approved):
🚀

PR #422, opened 2024-05-10
Plan_cartesian_motion uses Waypoints class

yck011522:
This is a short PR to migrate the plan_cartrsian_motion to use Waypoints class.

At the moment, only the FrameWaypoints are supported as no implementation is added for PointAxisWaypoints yet. Will do that in next PR.

- [] New feature in a **backwards-compatible** manner.

### Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
-* `Robot.plan_carteisan_motion()` now accepts `Waypoints` as target. Implementation for `FrameWaypoints` is supported with same functionality as before. Simply wrap `Frame` objects using `FrameWaypoints(frames)`.
+* `Robot.plan_cartesian_motion()` now accepts `Waypoints` as target. Implementation for `FrameWaypoints` is supported with same functionality as before. Simply wrap `Frame` objects using `FrameWaypoints(frames)`.
 * Changed `BoundingVolume`, `Constraint`, `JointConstraint`, `OrientationConstraint`, `PositionConstraint` to inherit from `compas.data.Data` class.
 * Change the signature of `plan_motion()` to use `target` (`Target` class) instead of `goal_constraints`. Only one target is accepted. Users who wish to compose their own constraint sets can still use `ConstraintSetTarget`.
 * Moved `Robot.orientation_constraint_from_frame()` to `OrientationConstraint.from_frame()`, as constraints are no longer intended for users to use directly.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+* `Robot.plan_carteisan_motion()` now accepts `Waypoints` as target. Implementation for `FrameWaypoints` is supported with same functionality as before. Simply wrap `Frame` objects using `FrameWaypoints(frames)`.
 * Changed `BoundingVolume`, `Constraint`, `JointConstraint`, `OrientationConstraint`, `PositionConstraint` to inherit from `compas.data.Data` class.
 * Change the signature of `plan_motion()` to use `target` (`Target` class) instead of `goal_constraints`. Only one target is accepted. Users who wish to compose their own constraint sets can still use `ConstraintSetTarget`.
 * Moved `Robot.orientation_constraint_from_frame()` to `OrientationConstraint.from_frame()`, as constraints are no longer intended for users to use directly.
--- a/docs/examples/02_description_models/03_targets.rst
+++ b/docs/examples/02_description_models/03_targets.rst
@@ -38,3 +38,10 @@ The :class:`compas_fab.robots.ConstraintSetTarget` class is used to specify a li
 constraints as a planning target. This is intended for advanced users who want to create custom
 combination of constraints. See :class:`compas_fab.robots.Constraint` for available
 constraints. At the moment, only the ROS MoveIt planning backend supports this target type.
+
+------------------------------------------
+Waypoints Target (Multiple Point Segments)
+------------------------------------------
+
+The :class:`compas_fab.robots.Waypoint` classes are used to describe a sequence of
+waypoints that the robot should pass through.
--- a/docs/examples/02_description_models/03_targets.rst
+++ b/docs/examples/02_description_models/03_targets.rst
@@ -38,6 +38,8 @@ constraints as a planning target. This is intended for advanced users who want t
 combination of constraints. See :class:`compas_fab.robots.Constraint` for available
 constraints. At the moment, only the ROS MoveIt planning backend supports this target type.
 
+.. _waypoints:
+
 ------------------------------------------
 Waypoints (Multiple Points / Segments)
 ------------------------------------------
--- a/docs/examples/02_description_models/03_targets.rst
+++ b/docs/examples/02_description_models/03_targets.rst
@@ -1,16 +1,15 @@
 .. _targets:
 
 *******************************************************************************
-Targets
+Targets and Waypoints
 *******************************************************************************
 
 -----------------------
-Single Targets (Static)
+Targets (Single Goal)
 -----------------------
 
 Target classes are used to describe the goal condition (i.e. end condition) of a robot
-for motion planning. They can be used for both Free Motion Planning (FMP) and Cartesian
-Motion Planning (CMP).
+for motion planning. They can be used for Free Motion Planning with :meth:`compas_fab.robots.Robot.plan_motion`.
 
 The :class:`compas_fab.robots.FrameTarget` is the most common target for motion planning.
 It defines the complete pose of the end-effector (or the robot flange, if no tool is attached).
@@ -40,8 +39,25 @@ combination of constraints. See :class:`compas_fab.robots.Constraint` for availa
 constraints. At the moment, only the ROS MoveIt planning backend supports this target type.
 
 ------------------------------------------
-Waypoints Target (Multiple Point Segments)
+Waypoints (Multiple Points / Segments)
 ------------------------------------------
 
-The :class:`compas_fab.robots.Waypoint` classes are used to describe a sequence of
-waypoints that the robot should pass through.
+The :class:`compas_fab.robots.Waypoints` classes are used to describe a sequence of
+waypoints that the robot should pass through in a planned motion. They are similar to Targets classes
+but contain a list of targets instead of a single target, which is useful for tasks such as
+drawing, welding or 3D printing.
+They can be used for Cartesian Motion Planning with :meth:`compas_fab.robots.Robot.plan_cartesian_motion`.
+
+The :class:`compas_fab.robots.FrameWaypoints` is the most common waypoint for Cartesian motion planning.
+It defines a list of complete pose for the end-effector (or the robot flange, if no tool is attached).
+It is created by a list of :class:`compas.geometry.Frame` objects or alternatively from a list of
+:class:`compas.geometry.Transformation` objects.
+
+The :class:`compas_fab.robots.PointAxisWaypoints` class is used for specifying a list of waypoints based on
+the Point-Axis concept used in the :class:`compas_fab.robots.PointAxisTarget`. Compared to
+:class:`~compas_fab.robots.FrameWaypoints`, this class allows for specifying a targets where the rotation
+around the Z-axis is not fixed. This is useful for example when the robot is using a cylindrical tool
+to perform a task, for example 3D printing, welding or drilling. The freely rotating axis is defined relative
+to the Z-axis of the tool coordinate frame (TCF). Note that the orientation of the tool
+at the end of the motion is not determined until after the motion is planned.
+
--- a/docs/examples/02_description_models/03_targets.rst
+++ b/docs/examples/02_description_models/03_targets.rst
@@ -9,7 +9,7 @@ Targets (Single Goal)
 -----------------------
 
 Target classes are used to describe the goal condition (i.e. end condition) of a robot
-for motion planning. They can be used for Free Motion Planning with :meth:`compas_fab.robots.Robot.plan_motion`.
+for motion planning. They can be used for Free-space Motion Planning with :meth:`compas_fab.robots.Robot.plan_motion`.
 
 The :class:`compas_fab.robots.FrameTarget` is the most common target for motion planning.
 It defines the complete pose of the end-effector (or the robot flange, if no tool is attached).
@@ -42,7 +42,7 @@ constraints. At the moment, only the ROS MoveIt planning backend supports this t
 Waypoints (Multiple Points / Segments)
 ------------------------------------------
 
-The :class:`compas_fab.robots.Waypoints` classes are used to describe a sequence of
+Waypoints classes are used to describe a sequence of
 waypoints that the robot should pass through in a planned motion. They are similar to Targets classes
 but contain a list of targets instead of a single target, which is useful for tasks such as
 drawing, welding or 3D printing.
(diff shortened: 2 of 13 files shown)


CHANGELOG.md
```diff
@@ -9,13 +9,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+* Added `compas_fab.robots.Waypoints` class to represent a sequence of targets. It has two child classes: `FrameWaypoints` and `PointAxisWaypoints`.
 * Added `compas_fab.robots.Target` class to represent a motion planning target.
 * Added also child classes `FrameTarget`, `PointAxisTarget`, `ConfigurationTarget`, `ConstraintSetTarget`
 * Unlike previous constraints, `Targets` do not contain `group` as parameter. Instead, group parameter is passed to the planning call.
 * Target scaling function is now embeded in the code for Targets. `scaled()` should be called by the user before passing the target to the `plan_motion` function.
 
 ### Changed
 
+* `Robot.plan_carteisan_motion()` now accepts `Waypoints` as target. Implementation for `FrameWaypoints` is supported with same functionality as before. Simply wrap `Frame` objects using `FrameWaypoints(frames)`.
```
gonzalocasas, line 20:
```suggestion
* `Robot.plan_cartesian_motion()` now accepts `Waypoints` as target. Implementation for `FrameWaypoints` is supported with same functionality as before. Simply wrap `Frame` objects using `FrameWaypoints(frames)`.
```

docs/examples/02_description_models/03_targets.rst
```diff
@@ -1,16 +1,15 @@
...
 
 -----------------------
-Single Targets (Static)
+Targets (Single Goal)
 -----------------------
 
 Target classes are used to describe the goal condition (i.e. end condition) of a robot
-for motion planning. They can be used for both Free Motion Planning (FMP) and Cartesian
-Motion Planning (CMP).
+for motion planning. They can be used for Free Motion Planning with :meth:`compas_fab.robots.Robot.plan_motion`.
```
gonzalocasas, line 12:
```suggestion
for motion planning. They can be used for Free-space Motion Planning with :meth:`compas_fab.robots.Robot.plan_motion`.
```

docs/examples/02_description_models/03_targets.rst
```diff
@@ -38,3 +37,27 @@ The :class:`compas_fab.robots.ConstraintSetTarget` class is used to specify a li
 constraints as a planning target. This is intended for advanced users who want to create custom
 combination of constraints. See :class:`compas_fab.robots.Constraint` for available
 constraints. At the moment, only the ROS MoveIt planning backend supports this target type.
+
+------------------------------------------
+Waypoints (Multiple Points / Segments)
+------------------------------------------
+
+The :class:`compas_fab.robots.Waypoints` classes are used to describe a sequence of
```
gonzalocasas, line 45:
```suggestion
Waypoints classes are used to describe a sequence of
```

src/compas_fab/backends/kinematics/analytical_plan_cartesian_motion.py
```diff
@@ -1,12 +1,25 @@
...
+
+from compas_fab.utilities import from_tcf_to_t0cf
+
+if not compas.IPY:
+    from typing import TYPE_CHECKING
+
+    if TYPE_CHECKING:
+        from typing import Optional  # noqa: F401
+        from compas_fab.robots import Robot  # noqa: F401
+        from compas_robots import Configuration  # noqa: F401
```
gonzalocasas, line 22: Please remove this from this PR. There's a separate PR for type hinting, and I would really like to address that concretely on the other one because I don't like this huge prefixing that all files will need to deal with ipy's lack of typing.

src/compas_fab/ghpython/components/Cf_PlanCartesianMotion/code.py
```diff
@@ -23,8 +24,9 @@ def RunScript(
 
         if robot and robot.client and robot.client.is_connected and start_configuration and planes and compute:
             frames = [plane_to_compas_frame(plane) for plane in planes]
+            frame_waypoints = FrameWaypoints(frames)
```
gonzalocasas, line 27: This got me thinking: this change is ok for this component, but how are we going to allow point-axis planning in Grasshopper, considering there's no point&axis entity that can be passed around? This is also evidenced by the fact that `PointAxisWaypoints` takes a list of tuples of points and vectors. So I first thought it could be a good idea to just pass a list of frames, and have a boolean, if that is set to true, we reduce the frame to a point&axis in the component. But there I started questioning if we need at all a `PointAxisWaypoints` class, or if we can instead have only `FrameWaypoints` containing that bool as part of its own data? It could be named something technical like `Underconstrained=True`, or something more humanly readable like `AllowRotationAroundNormal=True`. And since the Axis part of PointAxis is limited to the normal of the frame, I am currently leaning towards going with something more like the latter. What do you think?

src/compas_fab/robots/robot.py
```diff
@@ -1313,25 +1314,22 @@ def forward_kinematics(self, configuration, group=None, use_attached_tool_frame=
...
         """Calculate a cartesian motion path (linear in tool space).
 
         Parameters
         ----------
-        frames_WCF : :obj:`list` of :class:`compas.geometry.Frame`
-            The frames through which the path is defined.
+        waypoints : :class:`compas_fab.robots.Waypoints`
+            The waypoints for the robot to follow.
+            If a tool is attached to the robot, the :meth:`~compas_fab.robots.Waypoints.tool_coordinate_frame` parameter
+            should be set.
```
gonzalocasas, line 1325: I would rephrase this, I have a hard time parsing what it means.

src/compas_fab/robots/robot.py
```diff
@@ -1313,25 +1314,22 @@ def forward_kinematics(self, configuration, group=None, use_attached_tool_frame=
...
-            joints of the entire robot, at the starting position. Defaults to
-            the all-zero configuration.
+            joints of the entire robot, at the start of the motion.
+            Defaults to the all-zero configuration.
         group : :obj:`str`, optional
-            The planning group used for calculation. Defaults to the robot's
-            main planning group.
-        use_attached_tool_frame : :obj:`bool`, optional
-            If ``True`` and there is a tool attached to the planning group, it will use its TCF
-            instead of the T0CF to calculate cartesian paths. Defaults to ``True``.
```
gonzalocasas, line 1334: I'm not convinced with the removal of this. I think we need to review the idea of having `tool_coordinate_frame` in the targets and waypoints classes. It feels to me like these could contain the name or key of the tool to use, but the coordinate frame should not be attached inside them, otherwise, if we serialize that, and re-measure the tool, the waypoints need to be somehow retrofitted/updated with the new tcf. That doesn't make much sense.

src/compas_fab/robots/robot.py
```diff
@@ -1360,42 +1358,52 @@ def plan_cartesian_motion(
...
-            frames_WCF = self.from_tcf_to_t0cf(frames_WCF, group)
-
-        frames_WCF_scaled = []
-        for frame in frames_WCF:
-            frames_WCF_scaled.append(Frame(frame.point * 1.0 / self.scale_factor, frame.xaxis, frame.yaxis))
-
-        if path_constraints:
+        # Path constraints are only relevant to ROS Backend
+        path_constraints = options.get("path_constraints")
+        if need_scaling and path_constraints:
```
gonzalocasas, line 1406: This check is not correct. If options contains `path_constraints` but `need_scaling` ended up being false, we'll end up in the `else` block, setting ` path_constraints_WCF_scaled = None`, and basically ignoring them completely.

gonzalocasas: I don't know if adding this boolean flag to minimize scaling operations is so worth it. It makes the program logic harder to parse for humans, and it's kind of pointless because it's not a very expensive operation to do scale with a factor of 1.0 (which is what was happening before).

src/compas_fab/robots/targets.py
```diff
@@ -1,14 +1,25 @@
...
 from compas.data import Data
 from compas.geometry import Frame
 from compas.geometry import Transformation
 from compas_robots.model import Joint
 
+if not compas.IPY:
+    from typing import TYPE_CHECKING
+
+    if TYPE_CHECKING:
+        from compas_robots import Configuration  # noqa: F401
```
gonzalocasas, line 12: Same here, let's not mix things.

src/compas_fab/robots/targets.py
```diff
@@ -19,8 +30,8 @@ class Target(Data):
     pose, configuration, and joint constraints. Dynamic targets such as
     velocity, acceleration, and jerk are not yet supported.
 
-    Targets are intended to be used as arguments for the Backend's motion
-    planning methods. Different backends might support different types of
+    Waypoints are intended to be used for motion planning with a planning backend by using :meth:`compas_fab.robot.plan_motion`.
+    Different backends might support different types of
```
gonzalocasas, line 34: This is the wrong doc, it belongs to waypoints but it's on the target class.

tests/backends/kinematics/test_inverse_kinematics.py
```diff
@@ -95,22 +98,56 @@ def test_kinematics_client_with_attached_tool():
...
         options = {"solver": "ur5", "check_collision": True}
-        start_configuration = list(robot.iter_inverse_kinematics(frames_WCF[0], options=options))[-1]
-        trajectory = robot.plan_cartesian_motion(frames_WCF, start_configuration=start_configuration, options=options)
+
+        trajectory = robot.plan_cartesian_motion(frame_waypoints, options=options)
+
+        # Assert that the trajectory is complete
         assert trajectory.fraction == 1.0
+        # Assert that the trajectory has the correct number of points
+        assert len(trajectory.points) == len(frame_waypoints.target_frames)
```
gonzalocasas, line 153: This cannot be correct. The trajectory should most definitely contain more points that the number of input frames, if it's the same, the system didn't calculate anything except a series of IKs for the frames that were sent as input.

tests/backends/kinematics/test_inverse_kinematics.py
```diff
@@ -95,22 +98,56 @@ def test_kinematics_client_with_attached_tool():
...
+        # Assert that the trajectory is complete
+        assert trajectory.fraction == 1.0
+        # Assert that the trajectory has the correct number of points
+        assert len(trajectory.points) == len(frame_waypoints.target_frames)
+
+
+def test_kinematics_cartesian_with_tool_coordinate_frame(frame_waypoints):
+    if compas.IPY:
+        return
+    frame_waypoints.tool_coordinate_frame = Frame([0.01, 0.02, -0.03], [1, 0, 0], [0, 1, 0])
```
gonzalocasas, line 140: Again this feel incorrect. When seeing it here in a full example, it's clearer that is wrong imo. The tool needs to be defined with a key and a TCF, and the waypoints should only say e.g. `frame_waypoints.tool = "suction_gripper"`

tests/robots/test_targets.py
```diff
@@ -0,0 +1,224 @@
...
+    assert point_axis_waypoints.tool_coordinate_frame == nt.tool_coordinate_frame
+    assert point_axis_waypoints.name == nt.name
+
+
+def test_target_scale(frame_target):
+    scale_factor = 0.001
+    nt = frame_target.scaled(scale_factor)
+    assert nt.target_frame == frame_target.target_frame.scaled(scale_factor)
+    assert nt.tolerance_position == frame_target.tolerance_position * scale_factor
+    assert nt.tolerance_orientation == frame_target.tolerance_orientation * scale_factor
```
gonzalocasas, line 175: I don't know what happened here. This new file was added in another PR, why does it show up again here? Maybe you'd need to merge main back into this? Otherwise, this point has the bug that we fixed last Friday about scaling the orientation.

tests/backends/kinematics/test_inverse_kinematics.py
gonzalocasas: I see that there's a comment in the analytical solver saying that there's no interpolation going on. Even if we don't change that (which we should btw), testing explicitly for that behavior is not really helpful, because if we fix this for something better the tests will fail for no good reason.

src/compas_fab/backends/kinematics/analytical_plan_cartesian_motion.py
```diff
@@ -74,12 +115,23 @@ def plan_cartesian_motion(self, robot, frames_WCF, start_configuration=None, gro
         path = paths[idx]
         path = self.smooth_configurations(path)
         trajectory = JointTrajectory()
-        trajectory.fraction = len(path) / len(frames_RCF)
+        trajectory.fraction = len(path) / len(
+            frames_RCF
+        )  # Technically this always be 1.0 because otherwise, the path would be rejected earlier
```
gonzalocasas, line 120: Does black allow you to use such a strange line breaking?

src/compas_fab/backends/kinematics/analytical_plan_cartesian_motion.py
```diff
@@ -42,22 +54,51 @@ def plan_cartesian_motion(self, robot, frames_WCF, start_configuration=None, gro
...
         options.update({"keep_order": True})
 
+        # iterate over all input frames and calculate the inverse kinematics, no interpolation in between frames
         configurations_along_path = []
         for frame in frames_RCF:
             configurations = list(robot.iter_inverse_kinematics(frame, options=options))
             configurations_along_path.append(configurations)
 
+        # There is a maximum of 8 possible paths, corresponding to the 8 possible IK solutions for each frame
+        # The all() function is used to check if all configurations in a path are present.
```
gonzalocasas, line 101: Only true for 6DoF robots. It shouldn't be stated as a general truth.

src/compas_fab/backends/kinematics/analytical_plan_cartesian_motion.py
```diff
@@ -42,22 +54,51 @@ def plan_cartesian_motion(self, robot, frames_WCF, start_configuration=None, gro
...
+    def _plan_cartesian_motion_with_frame_waypoints(
+        self, robot, waypoints, start_configuration=None, group=None, options=None
+    ):
+        # type: (Robot, FrameWaypoints, Optional[Configuration], Optional[str], Optional[dict]) -> JointTrajectory
+        """Calculates a cartesian motion path with frame waypoints.
+
+        Planner behavior:
+        - If multiple paths are possible (i.e. due to multiple IK results), only the one that is closest to the start_configuration is returned.
+        - The path is checked to ensure that the joint values are continuous and that revolution values are the smallest possible.
+        - 'stepsize' is not used to sample in between frames (i.e. no interpolation), only the input frames are used.
```
gonzalocasas, line 78: I think it's called `max_step`, no?

gonzalocasas (changes requested):
I left a bunch of feedback, some of it might require some refactoring of the existing stuff.

src/compas_fab/ghpython/components/Cf_PlanCartesianMotion/code.py
gonzalocasas: Although I'm not 100% convinced that `PointAxis` still makes sense, I don't have a solid alternative to provide so, we'll keep it.

src/compas_fab/robots/targets.py
```diff
@@ -490,28 +490,30 @@ class Waypoints(Data):
...
         super(Waypoints, self).__init__()
         self.name = name
-
-    @property
-    def __data__(self):
-        raise NotImplementedError
+        # If the user provides a transformation, convert it to a Frame
+        if isinstance(tool_coordinate_frame, Transformation):
+            tool_coordinate_frame = Frame.from_transformation(tool_coordinate_frame)
+        self.tool_coordinate_frame = tool_coordinate_frame
```
gonzalocasas, line 516: This is problematic because it's redundant information: the tool has its frame defined, but this will repeat the same data, losing track of its source, so if the tool changes frame, the target has no way to know the data is incorrect.

gonzalocasas: #435

gonzalocasas (approved):
We're merging this, with the pending remark about #435 that really needs to be checked and decided on.

PR #438, opened 2024-06-20
Update Backends Documentation

yck011522:
Updated the docs for describing Backend Features interfaces. Note that the examples where how two planners can be combined is speculative to a future PR where the planner functions are called directly.

### DIFF
--- a/docs/developer/backends.rst
+++ b/docs/developer/backends.rst
@@ -43,10 +43,9 @@ For example:
             # insert fancy code here
             pass
 
-    class ExamplePlanner(
-        ExampleInverseKinematics,
-        PlannerInterface,
-    ):
+
+    # Now we can create a custom planner using our newly implemented backend feature
+    class ExamplePlanner(ExampleInverseKinematics, PlannerInterface):
         pass
 
 The planner can be instantiated and called in the following manner:
--- a/docs/developer/backends.rst
+++ b/docs/developer/backends.rst
@@ -30,7 +30,7 @@ For example:
 
     from compas.geometry import Frame
     from compas_fab.backends.interfaces import InverseKinematics
-    from compas_fab.backends.interfaces.client import PlannerInterface
+    from compas_fab.backends.interfaces import PlannerInterface
 
     class ExampleInverseKinematics(InverseKinematics):
         def inverse_kinematics(self, robot,
--- a/docs/developer/backends.rst
+++ b/docs/developer/backends.rst
@@ -19,18 +19,18 @@ but for now, such methods and attributes will be left with the
 client.
 
 The ``PlannerInterface`` serves as a template for any client-specific
-planner, providing default behavior for each of the
-methods listed within.  When a developer wishes to override any
-of these defaults, they should make use of the appropriate backend
-feature interface from ``backends/interfaces.py``.  The file
-``interfaces.py`` consists of a collection of classes, any
-implementation of which is callable through its ``__call__`` magic
-method.  For example:
+planner and contains all unified planning function signatures.
+Developers wishing to develop new planner features should inherit from the
+appropriate backend feature interface from ``backends/interfaces/backend_features.py``.
+These features are then composed into a single class that inherits from
+``PlannerInterface``.
+For example:
 
 .. code-block:: python
 
     from compas.geometry import Frame
     from compas_fab.backends.interfaces import InverseKinematics
+    from compas_fab.backends.interfaces.client import PlannerInterface
 
     class ExampleInverseKinematics(InverseKinematics):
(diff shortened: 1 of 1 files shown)


docs/developer/backends.rst
```diff
@@ -67,9 +76,16 @@ backend of ``ClientB`` is slow to compute inverse kinematics but can plan motion
...
+    with ClientA(robot) as client_a, ClientB(robot) as client_b:
+        inverse_kinematics = client_a.planner.inverse_kinematics
+        plan_motion = client_b.planner.plan_motion
+
+        frame = Frame([0, 0, 0], [1, 0, 0], [0, 1, 0])
+        ik_result = inverse_kinematics(frame)
+
+        start_configuration = robot.zero_configuration()
+        goal_configuration = robot.random_configuration()
+        motion_plan = plan_motion(start_configuration, goal_configuration)
```
gonzalocasas, line 88: This second example in general has never made much sense. I think we could kill the entire `ClientA`/`ClientB` example.

gonzalocasas (changes requested):
After looking quite lengthy at the current state, your diagrams, the various examples and so on, I think the `ClientInterface` as lost a bit its raison d'etre, and the relationship robot-client-planner is upside down for no good reason.

I propose the following:

- Client will no longer contain a `planner`, so all the forwarding of functions in `ClientInterface` goes away. Its responsibilites are mostly limited to just keeping a connection open to the external system.
- Planners will contain a `client` as today (but the relation will not be recursive as it is right now).
- `Robot` will not need a `client` anymore, instead it will need a `planner` instance in the constructor.

We discussed potentially removing the forwarding entirely from `Robot`, in which case, we could remove also `planner` from it completely.
However, this will have a huge impact on the API, and it affects all existing Grasshopper definitions that use COMPAS FAB quite a bit, so I say we keep the idea that `Robot` is the carrier of a connection to the planner by containing an instance of a planner.

This changes the relationship of the whole pipeline to be `Robot -> Planner(composite of backend features) -> Client`.

In code, this could look like this from an API perspective:

```python

# Example 1: ROS + MoveIt, explicit instances
client = RosClient()
robot = client.load_robot()
robot.planner = MoveItPlanner(client)

robot.inverse_kinematics(frame)


# Example 2: PyBullet, explicit instances
client = PyBulletClient()
robot = client.load_robot(urdf_filename)
robot.planner = PyBulletPlanner(client)

robot.inverse_kinematics(frame)

# Example 3: ROS + future Tesseract planner, explicit instances, @jf---
client = RosClient()
robot = client.load_robot()
robot.planner = TesseractPlanner(client)

robot.inverse_kinematics(frame)

# Example 4: ROS + custom planner
class SlerpInterpolatedPlanCartesianMotion(PlanCartesianMotion):
    def plan_cartesian_motion(self, *args, **kwargs):
        # here add nice implementation of interpolated cartesian planner
        pass

# This planner inherits all the features from MoveIt, and only replaces the cartesian motion planning feature with a custom one.
class CustomMoveItPlanner(SlerpInterpolatedPlanCartesianMotion, MoveItPlanner):
    pass

client = RosClient()
robot = client.load_robot()
robot.planner = CustomMoveItPlanner(client)

robot.plan_cartesian_motion(waypoints)

# Example 5: default planner creation via factory method
client = RosClient()
robot = client.load_robot()
robot.planner = client.new_planner()

robot.inverse_kinematics(frame)

# Example 6: non-default planner created via factory method (this is an alternative syntax for example 1, it's not mutually exclusive)
client = RosClient()
robot = client.load_robot()
robot.planner = client.new_planner(CustomMoveItPlanner)

robot.plan_cartesian_motion(waypoints)

```

docs/developer/backends.rst
yck011522: So do we remove the rest starting from "These interface also exist to allow mixing..."?

docs/developer/backends.rst
```diff
@@ -19,37 +19,46 @@ but for now, such methods and attributes will be left with the
...
+    from compas_fab.backends.interfaces.client import PlannerInterface
 
     class ExampleInverseKinematics(InverseKinematics):
         def inverse_kinematics(self, robot,
                                frame_WCF,
                                start_configuration=None,
                                group=None,
                                options=None):
+            # Access to the Client Interface is available:
+            client = self.client
```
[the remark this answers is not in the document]
yck011522: I'm not sure if `.is_connected` is implemented across all clients. For example an Analytical Planner will not really have that. Its also not in the interface.

gonzalocasas:
@yck011522 this is totally done, right? I'll close it, but if there's anything that is to recover from here, we can revive it

PR #442, opened 2024-10-13
Fix: handle exception in get_link_names for single-link group

### LINKED ISSUE #160: Robot's get_link_names throws exception for group with one link
**Describe the bug**
Robot's get_link_names throws exception for group with one link

**To Reproduce**
1. Load the ur5 demo robot via ROS.
2. Run the following code:
`
from compas_fab.backends import RosClient
from compas_fab.robots.ur5 import Robot

with RosClient() as client:
    robot = Robot(client)
    names = robot.get_link_names('endeffector')
`
**Expected behavior**
`names` should equal `['ee_link']`

**Actual behavior**
Exception is thrown "Exception: No chain found between the specified element"

 - OS: OSX
 - Python version 3.8
 - Python package manager conda

annapurna-gupta:
This pull request addresses an issue with the get_link_names method in the compas_fab library. Previously, this method threw an exception when called for groups containing only one link, indicating that no chain was found. The fix ensures that this method can now handle such cases without raising errors, allowing users to retrieve link names for single-link groups successfully.

- [X] Bug fix in a **backwards-compatible** manner.

##checklist

- [X] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [X] I ran all tests on my computer and it's all green (i.e. `invoke test`).
- [X] I ran lint on my computer and there are no errors (i.e. `invoke lint`).
- [X] I added new functions/classes and made them available on a second-level import, e.g. `compas_fab.robots.CollisionMesh`.
- [X] I have added tests that prove my fix is effective or that my feature works.


gonzalocasas (changes requested):
This does not properly fix the problem and it also has a few issues:
 - Please never commit your own virtual environment to a repository, you should remove everything from the `newenv` folder from this PR
 - The tests don't align with `pytest` conventions, a unit tests should not be just a script that you run, but a script that contains a function prefixed with `test_` and does unit testing on a specific feature. They also don't go into the `src` folder, they go into the `tests` folder
 - The fix is not really fixing the problem, you just added a `try..except` but re-raised the exception, so nothing has changed.

gonzalocasas:
I will need to change this. On one hand, the PR is outdated, but it also contains a large number of files that don't belong in the code tree.

PR #407, opened 2024-02-20
Robot library remove old robots

yck011522:
This depends on #406, we should review and merge #406 first before this one.

Removed `compas_fab.robots.ur5` because it is now part of `compas_fab.robots.RobotLibrary`.
Removed data files of ur5 and ur10e from `src/compas_fab/data/universal_robots` because they are now in of `src/compas_fab/data/robot_library`.

Future constructs should use:
```
from compas_fab.robots import RobotLibrary
robot = RobotLibrary.ur5(load_geometry=True)
```
Or from URDF and SRDF from scratch
```
urdf_filename = compas_fab.get("robot_library/ur5_robot/urdf/robot_description.urdf")
robot_model = RobotModel.from_urdf_file(urdf_filename )

mesh_loader = LocalPackageMeshLoader(compas_fab.get("robot_library/ur5_robot"), "")
robot_model.load_geometry(mesh_loader)

srdf_filename = compas_fab.get("robot_library/ur5_robot/robot_description_semantic.srdf")
semantics = RobotSemantics.from_srdf_file(srdf_filename, robot_model)
robot = Robot(robot_model, semantics=semantics)
```

### What type of change is this?

- [ ] Bug fix in a **backwards-compatible** manner.
- [ ] New feature in a **backwards-compatible** manner.
- [x] Breaking change: bug fix or new feature that involve incompatible API changes.
- [ ] Other (e.g. doc update, configuration, etc)

### Checklist

_Put an `x` in the boxes that apply. You can also fill these out after creating the PR. If you're unsure about any of them, don't hesitate to ask. We're here to help! This is simply a reminder of what we are going to look for before merging your code._

- [ ] I added a line to the `CHANGELOG.md` file in the `Unreleased` section under the most fitting heading (e.g. `Added`, `Changed`, `Removed`).
- [ ] I ran all tests on my computer and it's all green (i.e. `invoke test`).
- [ ] I ran lint on my computer and there are no errors (i.e. `invoke lint`).
- [ ] I added new functions/classes and made them available on a second-level import, e.g. `compas_fab.robots.CollisionMesh`.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added necessary documentation (if appropriate)

PR #444, opened 2025-03-24
Gh cpython -> `LTS-main-1.x`

chenkasirer:
* Added CPython components.
* Added some helper functions in `compas_fab.ghpython.components`.
* Added the yakerize workflow.
* Updated dependency to `compas_invocations2`.
* Fixed small bug in `inverse_kinematics_spherical_wrist`.

### Checklist


src/compas_fab/backends/kinematics/solvers/spherical_wrist.py
```diff
@@ -174,7 +174,10 @@ def inverse_kinematics_spherical_wrist(target_frame, points):
         elbow_frame = Frame(p1A, elbow_dir, [0, 0, 1])
         elbow_plane = (p1A, elbow_frame.normal)
 
-        _, (center, radius, normal) = intersection_sphere_sphere(sphere1, sphere2)
+        sphere1_prim = (sphere1.base, sphere1.radius)
+        sphere2_prim = (sphere2.base, sphere2.radius)
+
+        _, (center, radius, normal) = intersection_sphere_sphere(sphere1_prim, sphere2_prim)
```
gonzalocasas, line 180: Is this due to a change in compas? Do we need to correct/limit the required version?

src/compas_fab/ghpython/components/__init__.py
```diff
@@ -1,5 +1,62 @@
 from __future__ import absolute_import
 
+try:
+    import Grasshopper  # type: ignore
+except ImportError:
+    pass
+
 
 def create_id(component, name):
     return "{}_{}".format(name, component.InstanceGuid)
```
gonzalocasas: This method is actually now on [`compas_ghpython`](https://github.com/compas-dev/compas/blob/main/src/compas_ghpython/__init__.py), so we could remove and replace for that one.

gonzalocasas (approved):
Amazing! Thank you so much! I added some minor comment, but it lgtm

src/compas_fab/backends/kinematics/solvers/spherical_wrist.py
chenkasirer: yes but it's already set properly. `__iter__` got removed from a bunch of shapes in COMPAS2

gonzalocasas: 👍

chenkasirer:
@gonzalocasas any idea what's up with this integration test? got it to run locally now but not on remote. seems it hasn't been working for while

gonzalocasas:
> @gonzalocasas any idea what's up with this integration test? got it to run locally now but not on remote. seems it hasn't been working for while

Maybe we just disable these failing tests. @yck011522 tried a ton of things to fix those, and I think the solution in the end was to skip them: https://github.com/compas-dev/compas_fab/pull/407/commits/9229af342bee14bab7afdedc131f58679069414c

[image
](https://github.com/compas-dev/compas_fab/pull/407)

yck011522:
The Ros connection thing occasionally work if you just re run it. 

Bass on my previous experience, I superstitiously suspect that The py test that is performed in remote can sometimes run multiple tests at the same time, resulting in multiple connections to the same Ros instance which cause it to crash. 

I also have a feeling that this affects tests to inline codes inside RST files. For proper tests that are inside its independent python file this never seems to be a problem.

chenkasirer:
ugh I give up. made some fixes to make ironpython tests run but they will still fail until https://github.com/compas-dev/compas/pull/1454 is merged.
integration tests I dunno, skipped some as suggested but others still fail. @gonzalocasas @yck011522  are you guys good with merging this PR and we deal with this in a future one?

PR #460, opened 2026-06-10
docs: add robot scaling examples for issue #166

### LINKED ISSUE #166: Add examples for scaling a robot.
# Feature Request

It would be nice to have an example showing how to load a robot from URDF / ROS (robot in meters) and then scale it (for millimeters).

```
robot = client.load_robot(load_geometry=True)
robot.artist = RobotArtist(robot.model)
robot.scale(1000)
```

It is not clear to me whether scale() should be before setting artist or vise versa. In my opinion, both order should work but currently only my example will work.

It is also not clear to me how the scaling affects path planning input and output. Are those values also scaled or are they in native robot units?

It is also not clear to be whether scale() is a modal setting (setting a parameter) or an action (like a transformation). For example, the behavior of repeatedly calling robot.scale is not documented. I personally think it should behave as a modal setting to avoid potential issues.

Parsa-Meshkini:
## Summary

Closes #166

Adds documentation and runnable examples showing how to scale a robot model from meters (URDF/ROS default) to millimeters.

**New files:**
- `docs/examples/02_description_models/04_scaling_robot.rst` — documentation page covering the scaling workflow
- `docs/examples/02_description_models/files/04_scaling_robot.py` — standalone example using `RobotLibrary` (no ROS required)
- `docs/examples/02_description_models/files/04_scaling_robot_from_ros.py` — example loading from ROS and scaling after setting the scene object

**What the docs explain:**
- `scale()` applies an absolute (non-cumulative) factor — calling it twice with `1000` stays at `1000x`
- `scale()` can be called before or after assigning `scene_object`; both work correctly
- Affects all planning I/O (targets, trajectories) automatically

## Test plan

- [x] `04_scaling_robot.py` runs without errors (`PYTHONPATH=src venv/bin/python docs/examples/02_description_models/files/04_scaling_robot.py`)
- [x] Output confirms correct scale factor at each step
- [ ] ROS example requires a live ROS backend (consistent with other `*_from_ros.py` examples in this folder)

### DIFF
--- a/docs/examples/02_description_models/04_scaling_robot.rst
+++ b/docs/examples/02_description_models/04_scaling_robot.rst
@@ -1,0 +1,75 @@
+.. _scaling_robots:
+
+*******************************************************************************
+Scaling robots
+*******************************************************************************
+
+Robot models loaded from URDF files — whether from disk, Github, or a running
+ROS instance — follow the `ROS REP-0103 <https://www.ros.org/reps/rep-0103.html>`_
+convention and store all geometry in **meters**. Many CAD environments (Rhino,
+Grasshopper, etc.) work in **millimeters** by default. The
+:meth:`~compas_fab.robots.Robot.scale` method bridges this gap.
+
+Scaling a robot model
+=====================
+
+The simplest way to scale is to load a robot and call
+:meth:`~compas_fab.robots.Robot.scale` with the desired multiplier.
+
+.. literalinclude :: files/04_scaling_robot.py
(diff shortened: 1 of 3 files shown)


gonzalocasas (changes requested):
@Parsa-Meshkini thanks for the contribution and apologies for the late reply. This PR would need to be rebased and updated to the latest `main`. When you sent it, we were in the middle of a large refactoring.

PR #465, opened 2026-06-24
Fix get_link_names exception for single-link groups (#160)

### LINKED ISSUE #160: Robot's get_link_names throws exception for group with one link

Swastik-Verma:
Fixes #160

## Problem
`RobotCell.get_link_names()` raised an exception when called on a group
consisting of a single link (e.g. an end-effector group with no
configurable joints), instead of returning that link's name.

## Root cause
When a group's base link and end-effector link are the same (i.e. the
group has only one link), `get_link_names()` would still call
`iter_link_chain()`, which internally tries to compute a shortest path
between two identical nodes. This returns an empty path, which was
incorrectly treated as "no chain found," raising an exception even
though the group is valid.

## Fix
Added a check in `get_link_names()`: if the base link and end-effector
link are the same, the function returns that single link's name
directly, without going through the chain-finding logic.

## Testing
Added a test case in `test_get_link_names` using the `endeffector`
group on the `ur10e_gripper_one_beam` fixture, which has a single link
and zero configurable joints — reproducing the original bug scenario.
Verified the fix with:
- `invoke test --doctest`
- `invoke testcodeblocks`

Both show the same pre-existing, unrelated baseline failures as before
this change (documented in issue discussion/PR comments if needed),
with no new failures introduced.

### DIFF
--- a/tests/robots/test_robot_cell.py
+++ b/tests/robots/test_robot_cell.py
@@ -178,6 +178,9 @@ def test_get_link_names(panda, rfl, ur10e_gripper_one_beam, abb_irb4600_40_255_g
     _test(*rfl)
     _test(*ur10e_gripper_one_beam)
     _test(*abb_irb4600_40_255_gripper_one_beam)
+    robot_cell, robot_cell_state = ur10e_gripper_one_beam
+    link_names = robot_cell.get_link_names("endeffector")
+    assert link_names == ["tool0"]
 
 
 def test_default_touch_links_returns_first_geometry_bearing_link():
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -17,3 +17,4 @@
 - Edvard Bruun <<ebruun@princeton.edu>> [@ebruun](https://github.com/ebruun)
 - Victor Pok Yin Leung <<leung@arch.ethz.ch>> [@yck011522](https://github.com/yck011522)
 - Begüm Saral <<begum.saral@tum.de>> [@begums](https://github.com/begums)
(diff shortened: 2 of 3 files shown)


tests/robots/test_robot_cell.py
```diff
@@ -178,6 +178,9 @@ def _test(rc: RobotCell, rcs: RobotCellState):
     _test(*rfl)
     _test(*ur10e_gripper_one_beam)
     _test(*abb_irb4600_40_255_gripper_one_beam)
+    robot_cell, robot_cell_state = ur10e_gripper_one_beam
+    link_names = robot_cell.get_link_names("endeffector")
+    assert link_names == ["tool0"]
```
gonzalocasas, line 183: This doesn't really test that the fix works. You'd need to add a test that has a robot with a single link group.

gonzalocasas (changes requested):
Thanks for the contribution! Please add a line to the changelog and check the feedback regarding the test!

tests/robots/test_robot_cell.py
Swastik-Verma: Thanks for the feedback! I'm a bit busy right now but will address both points soon.

PR #477, opened 2026-08-08
feat: make Waypoints behave like a list (fixes #430)

### LINKED ISSUE #430: Make FrameWaypoints and PointAxisWaypoints behave like a list
So we can do something like 
```
waypoint = FrameWaypoint(...)
waypoint.append(...)
```

ChickenisLegit:
Closes #430. Added list-like methods to the Waypoints base class so that FrameWaypoints and PointAxisWaypoints can be iterated, indexed, appended to, etc.

### DIFF
--- a/.github/workflows/pr-checks.yml
+++ b/.github/workflows/pr-checks.yml
@@ -13,9 +13,8 @@ jobs:
     steps:
       - uses: actions/checkout@v6
       - name: Changelog check
-        uses: Zomzog/changelog-checker@v1.1.0
+        uses: Zomzog/changelog-checker@v1.3.0
         with:
           fileName: CHANGELOG.md
-          checkNotification: Simple
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


gonzalocasas:
@ChickenisLegit thanks for the contribution!!! 🌟  Could you please add a bullet point in the changelog file describing the change and a little unit test for the feature?

ChickenisLegit:
@gonzalocasas Done! Added the changelog entry and the unit tests.

yck011522:
Hi @ChickenisLegit thanks for the contribution. Look good to me. 

Just a stylistic suggestion, would you consider moving the implementation of the `waypoints` property into the two child class. That way whoever implementing new child classes would not need to fiddle with the base class. 

You can choose to merge as is or choose to move it and merge.

yck011522 (approved):

gonzalocasas:
@yck011522 very good point!

@ChickenisLegit could you please move it to the subclasses and then we can merge?

Thanks!

yck011522 (approved):
Looks good

PR #488, opened 2026-09-08
Fix ABB/RFL inertias and preserve PyBullet base-link placement

### LINKED ISSUE #486: tensors in the bundled robot library are not physically realisable (MuJoCo refuses to load them)
The `<inertial>` blocks in two of the bundled robot descriptions describe bodies
that cannot exist. This is not a precision complaint -- the tensors are
rejected outright by physics engines that validate them, and silently discarded
by ones that do not.

## What is wrong

A rigid body's inertia tensor must be symmetric **positive definite**, and its
eigenvalues (the principal moments) must additionally satisfy the triangle
inequality `J_a + J_b >= J_c`. The second condition follows from each principal
moment being an integral of a sum of two squared coordinates, so no arrangement
of matter can violate it. A tensor can be positive definite and still describe
no body.

Both files fail:

| file | tensors | invalid |
|---|---|---|
| `abb_irb4600_40_255/urdf/robot_description.urdf` | 7 | **7** |
| `rfl/urdf/robot_description.urdf` | 38 | **38** |

Per link in the ABB file, eigenvalues computed with `numpy.linalg.eigvalsh`:

| link | principal moments | fails |
|---|---|---|
| `base_link` (L20) | `[-6.0413, -5.9520, +1.8893]` | not positive definite |
| `link_1` (L40) | `[-6.0413, -5.9520, +1.8893]` | not positive definite |
| `link_2` (L60) | `[-20.3664, -18.8855, +2.6159]` | not positive definite |
| `link_3` (L80) | `[+1.2497, +2.5280, +4.9783]` | positive definite, **triangle short by 24.1%** |
| `link_4` (L100) | `[-8.0000, -8.0000, +0.4820]` | not positive definite |
| `link_5` (L120) | `[-0.0090, -0.0040, +0.0400]` | not positive definite |
| `link_6` (L140) | `[-0.3293, -0.3264, +0.0251]` | not positive definite |

## It is one defect, copy-pasted

The 45 bad tensors are only **6 distinct values**. In `rfl`, the single block

```xml
<inertia ixx="-4.98" ixy="-0.0" ixz="2.7" iyy="-5.952" iyz="-0.0" izz="0.828"/>
```

appears **18 times**, plus twice more in the ABB file -- 20 links sharing one
tensor. The remaining five values appear 4 times each. So this is one bad
inertial block propagated across the library, not 45 independent mistakes.

Two other signs that the whole block is placeholder data rather than CAD output:
`base_link`, `link_1`, `link_2` and `link_3` all declare **exactly 120.0 kg**,
and several inertial origins are round numbers far from the link's actual
geometry (`link_4` declares `xyz="0 0 1"` while its own collision mesh has its
centroid at `[0.654, 0.001, -0.001]`).

## Measured consequence

MuJoCo 3.11.0, loading the ABB file as shipped (`package://` resolved locally so
the meshes are found):

```
Error: error 'inertia must have positive eigenvalues' in fullinertia
```

The model does not load at all. PyBullet takes the other path -- it accepts the
file and silently zeroes the offending tensors, so a pipeline built on it runs
for years with massless links and never reports anything.

I appreciate compas_fab is a planning library and these values may simply never
be read on your own code paths. The reason it is still worth fixing is that the
bundled URDFs get handed to other tools, and in the strict half of them the
robot does not load.

## Proposed correction

Since the values are placeholders, there is no CAD ground truth to restore. But
the repo already ships the geometry, so a physically valid tensor can be derived
from **each link's own collision mesh at its own declared mass**, assuming
uniform density:

| link | principal moments (mesh-derived) | valid |
|---|---|---|
| `base_link` | `[1.9529, 4.0144, 5.2338]` | yes |
| `link_1` | `[3.2677, 4.0876, 5.0694]` | yes |
| `link_2` | `[0.8133, 14.4835, 14.8137]` | yes |
| `link_3` | `[3.5073, 4.4812, 5.1427]` | yes |
| `link_4` | `[0.1924, 3.7274, 3.7654]` | yes |
| `link_5` | `[0.0299, 0.0299, 0.0313]` | yes |
| `link_6` | `[0.0052, 0.0052, 0.0074]` | yes |

The inertial origins are set to the same meshes' centroids, since leaving the
declared origins in place would put physically valid tensors at positions the
geometry contradicts.

**Verified after patching:** MuJoCo loads the model (`nbody=7`, total mass
415 kg -- 535 kg declared, less the 120 kg base that becomes the world body) and
runs 500 steps with all state finite.

**Honest limits of this fix.** Uniform density is an approximation: a real
IRB4600 concentrates mass in the joint motors and gearboxes, so these numbers
are not what ABB's CAD would give. They are *physically realisable* values
consistent with the geometry and masses the repo already declares, which the
current ones are not. If you have access to real values, those should win. The
integrity tool used here is a straightforward signed-tetrahedron mesh
integration, self-tested against a closed box's analytic `m*(y^2+z^2)/12` to
2.8e-17.

The patch below covers the ABB file only. `rfl` needs the same treatment; I held
off because its links are assembled differently and I did not want to guess at
which meshes back which links -- happy to do it if the approach looks right to
you.

<details>
<summary><b>Patch</b> for <code>abb_irb4600_40_255</code> (<code>git apply</code>-able, 14 lines changed)</summary>

```diff
--- a/src/compas_fab/data/robot_library/abb_irb4600_40_255/urdf/robot_description.urdf
+++ b/src/compas_fab/data/robot_library/abb_irb4600_40_255/urdf/robot_description.urdf
@@ -15,9 +15,9 @@
       </geometry>
     </collision>
     <inertial>
-      <origin xyz="0.0 0.0 0.0" rpy="0.0 0.0 0.0"/>
+      <origin xyz="-0.103748 0.000230 0.090733" rpy="0.0 0.0 0.0"/>
       <mass value="120.0"/>
-      <inertia ixx="-4.98" ixy="-0.0" ixz="2.7" iyy="-5.952" iyz="-0.0" izz="0.828"/>
+      <inertia ixx="1.95288" ixy="-0.00131035" ixz="0.230948" iyy="4.01444" iyz="0.00106956" izz="5.23378"/>
     </inertial>
   </link>
   <link name="link_1">
@@ -35,9 +35,9 @@
       </geometry>
     </collision>
     <inertial>
-      <origin xyz="0.18 0.0 0.5" rpy="0.0 0.0 0.0"/>
+      <origin xyz="0.051430 0.023260 0.403788" rpy="0.0 0.0 0.0"/>
       <mass value="120.0"/>
-      <inertia ixx="-4.98" ixy="-0.0" ixz="2.7" iyy="-5.952" iyz="-0.0" izz="0.828"/>
+      <inertia ixx="3.26767" ixy="-0.282499" ixz="-0.564478" iyy="5.06942" iyz="-0.208255" izz="4.0876"/>
     </inertial>
   </link>
   <link name="link_2">
@@ -55,9 +55,9 @@
       </geometry>
     </collision>
     <inertial>
-      <origin xyz="0.94 0.08 -0.34" rpy="0.0 0.0 0.0"/>
+      <origin xyz="-0.000000 -0.180953 0.488128" rpy="0.0 0.0 0.0"/>
       <mass value="120.0"/>
-      <inertia ixx="-2.96" ixy="2.256" ixz="-9.588" iyy="-18.72601" iyz="-0.816" izz="-14.95"/>
+      <inertia ixx="14.4835" ixy="2.59188e-09" ixz="4.30393e-08" iyy="14.8137" iyz="0.282259" izz="0.813278"/>
     </inertial>
   </link>
   <link name="link_3">
@@ -75,9 +75,9 @@
       </geometry>
     </collision>
     <inertial>
-      <origin xyz="0.07 -0.266 0.088" rpy="0.0 0.0 0.0"/>
+      <origin xyz="0.042498 0.043936 0.119882" rpy="0.0 0.0 0.0"/>
       <mass value="120.0"/>
-      <inertia ixx="1.345" ixy="-0.5586" ixz="0.1848" iyy="4.65068" iyz="-0.70224" izz="2.76032"/>
+      <inertia ixx="3.51955" ixy="0.100722" ixz="-0.0719538" iyy="5.1305" iyz="0.067088" izz="4.48114"/>
     </inertial>
   </link>
   <link name="link_4">
@@ -95,9 +95,9 @@
       </geometry>
     </collision>
     <inertial>
-      <origin xyz="0.0 0.0 1.0" rpy="0.0 0.0 0.0"/>
+      <origin xyz="0.654158 0.001138 -0.001139" rpy="0.0 0.0 0.0"/>
       <mass value="40.0"/>
-      <inertia ixx="-8.0" ixy="-0.0" ixz="-0.0" iyy="-8.0" iyz="-0.0" izz="0.482"/>
+      <inertia ixx="0.192446" ixy="-0.0118404" ixz="0.00589281" iyy="3.72732" iyz="-8.8271e-06" izz="3.76544"/>
     </inertial>
   </link>
   <link name="link_5">
@@ -115,9 +115,9 @@
       </geometry>
     </collision>
     <inertial>
-      <origin xyz="0.14 0.0 0.0" rpy="0.0 0.0 0.0"/>
+      <origin xyz="-0.000000 -0.001500 0.000000" rpy="0.0 0.0 0.0"/>
       <mass value="10.0"/>
-      <inertia ixx="0.04" ixy="-0.0" ixz="-0.0" iyy="-0.009" iyz="-0.0" izz="-0.004"/>
+      <inertia ixx="0.0299416" ixy="-1.41878e-20" ixz="6.42562e-19" iyy="0.0312815" iyz="-1.96855e-19" izz="0.0299416"/>
     </inertial>
   </link>
   <link name="link_6">
@@ -135,9 +135,9 @@
       </geometry>
     </collision>
     <inertial>
-      <origin xyz="0.081 -0.091 0.332" rpy="0.0 0.0 0.0"/>
+      <origin xyz="-0.000000 0.000000 -0.019753" rpy="0.0 0.0 0.0"/>
       <mass value="5.0"/>
-      <inertia ixx="-0.29652" ixy="-0.03686" ixz="0.09396" iyy="-0.28792" iyz="-0.10556" izz="-0.04621"/>
+      <inertia ixx="0.00517999" ixy="7.79558e-08" ixz="-1.44659e-20" iyy="0.00518171" iyz="-6.8487e-20" izz="0.00742259"/>
     </inertial>
   </link>
   <link name="tool0"/>
```

</details>

gonzalocasas:
The bundled ABB IRB4600-40/2.55 and RFL URDFs contain 45 physically invalid inertia tensors, which prevent MuJoCo from loading either model. Apply the two patches [contributed by @omnilink-tech](https://github.com/compas-dev/compas_fab/issues/486#issuecomment-5454028524), correcting the tensors and centres of mass for all 7 ABB and 38 RFL inertial links.

Applying the URDF changes also exposed a PyBullet integration bug: setting the ABB base to the world origin displaced its tool by about 14 cm because the backend passed a base-link pose to an API that expects the base's inertial pose. Convert between those frames when reading or setting a base pose, including object placement. Regression tests cover translated and rotated inertial frames at two base poses and check the child-link pose as well as the base, so matching setter/getter errors cannot hide a displacement.

The replacement values are uniform-density approximations of the collision meshes at the existing declared masses, **not manufacturer dynamic parameters**. The masses remain unvalidated: their implied densities range from about 1.08 kg/m³ for the RFL gantry to 14,171 kg/m³ for the ABB wrist. Both URDFs now carry this caveat. The numerical patch values are unchanged; only explanatory XML comments were added.

Closes #486.

### Independent validation

- Recomputed all 45 centres of mass and tensors with Trimesh 4.12.2, independently of the contributor's generator. All source collision meshes are watertight with consistent winding and positive volume. Applied mesh scales and collision-origin transforms, including the RFL gantry translation and rotated components. Maximum centre discrepancy: `4.19e-8 m`; maximum relative tensor error (Frobenius norm): `3.54e-9`. Checked the integration library against the analytic inertia of a rotated and translated box.
- Verified that masses, inertial orientations, joints, collision geometry, and all other XML data are unchanged. Only inertial positions and tensors differ from `main`.
- Added a NumPy regression test for positive definiteness and the principal-moment triangle inequality across all **94 tensors in 11 bundled URDFs**. Confirmed that it rejects the original ABB and RFL files.
- MuJoCo **3.12.0**, with package paths resolved locally and `inertiafromgeom="false"`, `balanceinertia="false"`: both originals fail with non-positive inertia eigenvalues; both corrected models load and run with finite positions, velocities and accelerations, no warnings, and no simulation resets. ABB: 7 bodies, 415 kg, 500 steps. RFL: 35 bodies, 3340 kg, 5000 steps. These are smoke checks, not validation of real-machine dynamics.
- Confirmed the four new PyBullet regression cases fail before the backend fix and pass afterward; the existing ABB forward-kinematics test also passes afterward.
- `pytest -q`: **195 passed, 21 skipped** (opt-in live ROS integration tests).
- `invoke lint`, Ruff formatting checks for changed Python files, and `git diff --check`: passed.

### Checklist

- [x] Added changelog entries under `Unreleased`.
- [x] Ran the local test suite; live ROS integration tests were skipped as noted above.
- [x] Ran lint with no errors.
- [x] Added regression tests.
- [x] Documented the approximation in the affected URDFs and credited the contributor in `AUTHORS.md`.


gonzalocasas:
@yijiangh @yck011522 hey guys, this bug report surfaced something unexpected in the pybullet client. It looks rather decent to me, but maybe you can take a deeper look at it? Thanks!!!

